PR #4, opened 2023-02-06
Improve docs and github actions

ausias-armesto:
- Added documentation on how to run the Image
- Improved the way that the image is being built and pushed

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - name: Build and push
+      - name: Build
         uses: docker/build-push-action@v4
         with:
           push: false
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -1,43 +1,24 @@
...
           fetch-depth: 0
-
-      - name: Set up Google Cloud Credentials
-        uses: google-github-actions/auth@v0
-        with:
-          credentials_json: ${{ secrets.GOOGLE_HOPRASSOCIATION_CREDENTIALS_REGISTRY }}
-
-      - name: Set up Google Cloud SDK
-        uses: google-github-actions/setup-gcloud@v0
+      - name: Build and push
```
Teebor-Choka, line 17: Push is set to `false`, so the step name should probably by only `Build`:
```suggestion
      - name: Build
```

ausias-armesto: Thanks, issues with Copy&Paste ;)

Teebor-Choka (approved):

PR #14, opened 2023-05-29
Web3 and Safe initial redux structure

mjadach-iv:
- adding structure for web3 and safe in Redux
 - adding wagmi updater using hooks to update redux store
 - adding modal to connect to metamask


src/providers/Wagmi/updater.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+// wagmi
+import { useAccount, useConnect, useDisconnect } from 'wagmi';
+
+// Redux
+import { useAppDispatch, useAppSelector } from '../../store';
+import { Store } from '../../types/index';
+import { web3Actions } from '../../store/slices/web3';
+
+export default function WagmiProvider() {
```
diegoalzate, line 11: maybe we should call this WagmiUpdater as there is already a WagmiProvider

mjadach-iv: Good point, changed

src/components/ConnectWeb3/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  gap: 8px;
+  margin-top: 8px;
+  p {
+    margin-top: 48px;
+  }
+`;
+
+export default function ConnectWeb3() {
+  const dispatch = useAppDispatch();
+  const [chooseWalletModal, set_chooseWalletModal] = useState(false);
```
diegoalzate, line 26: `set_chooseWalletModal` are we going to use this convention? I think this can look a bit out of place in our repo but if we do this then all are sets have to be like this.

src/sections/selectNode.tsx
```diff
@@ -43,14 +43,18 @@ function Section1() {
     dispatch(authActions.useNodeData({ ip, apiKey }));
     dispatch(authActions.addNodeData({ ip, apiKey: saveApiKey ? apiKey : '' }));
     dispatch(authActionsAsync.login({ ip, apiKey }));
-    dispatch(nodeActionsAsync.getInfo());
+    dispatch(
+      nodeActionsAsync.getInfoThunk({ apiToken: apiKey, apiEndpoint: ip })
```
diegoalzate, line 47: should rename local state here so it does not handle the variables with name: `apiKey` and `ip`

src/store/slices/auth/actionsAsync.ts
```diff
@@ -1,11 +1,12 @@
 import { authActions } from './index';
-import { api } from 'hopr-sdk';
+import { api } from '@hoprnet/hopr-sdk';
+const { getInfo } = api;
 
 export const login = (loginData: any) => {
```
diegoalzate, line 5: this should receive a type and I think it would be something along the lines of {apiToken: string; apiEndpoint: string}

src/store/slices/node/actionsAsync.ts
```diff
@@ -603,7 +606,7 @@ export const createExtraReducers = (
       const { balance, channelId, peerId, status, type } = action.payload;
       // find channel if it already exists
       const channelIndex = state.channels?.[type].findIndex(
-        (channel) => channel.channelId === channelId
+        (channel: any) => channel.channelId === channelId
```
diegoalzate, line 609: is this :any needed? i would assume it i inferred from state.channels type

src/store/slices/node/actionsAsync.ts
```diff
@@ -642,7 +645,7 @@ export const createExtraReducers = (
         // using challenge as an id between tickets
         const uniqueIdentifier = updatedTicket.challenge;
         const existingIndex = state.tickets?.findIndex(
-          (ticket) => ticket.challenge === uniqueIdentifier
+          (ticket: any) => ticket.challenge === uniqueIdentifier
```
diegoalzate, line 648: same as comment above on typing

src/components/ConnectWeb3/index.tsx
mjadach-iv: Yes, this way it is a lot easier to find what you are looking for with case sensitive search, and also you can simply copy pase the const name with addition of _ instead of changing the 1st letter to caps

src/sections/selectNode.tsx
mjadach-iv: Good point, changed

src/store/slices/auth/actionsAsync.ts
mjadach-iv: Good point, changed

src/store/slices/node/actionsAsync.ts
mjadach-iv: Good point, changed

src/store/slices/node/actionsAsync.ts
mjadach-iv: Good point, changed

diegoalzate (approved):

PR #25, opened 2023-05-30
Ws connection middleware for redux

### LINKED ISSUE #19: Admin UI: link websocket for receiving messages to the store
Definition of done:
 - work to be done only in Redux or app, no html
 - each message should be stored in redux
 - each message should have a property 'seen' and 'timestamp'
 - there should be a counter of unseen messages in redux

diegoalzate:
closes #19 

### How does this work? 
[Redux middleware](https://redux.js.org/understanding/history-and-design/middleware#the-final-approach) it provides a third-party extension point between dispatching an action, and the moment it reaches the reducer. So when a initializeWebsocket action is triggered it will first run run through the middleware creating a websocket connection before reaching the actual reducer.

### Why?
The WebSocket instance is not serializable and thus technically not something that should be stored in Redux state. So there were 2 alternatives considered in this pr:

#### [rtk query streaming updates](https://redux-toolkit.js.org/rtk-query/usage/streaming-updates)
With this solution we could have an easy way to mount ws connection and tear it down, however there are a lot of concepts the team would need to learn and it also provides less customization. 

#### custom redux middleware
Even though this solution is more code, the concept of middleware is easier to understand and provides us a solution that can adapter better to any changes that may happen on hoprd side.

### why was LocalRootState created?
The middleware has access to the whole store, however I could not reference the store type inside the actual middleware because the middleware will be used in the store. That would create a circular dependency. We could create the RootStore type by hand instead of inferring but I think the store is more likely to change than the middleware.

### How can this be better?
- add common ws logic to hopr-sdk and use it here

### DIFF
--- a/src/store/slices/node/websocketMiddleware.ts
+++ b/src/store/slices/node/websocketMiddleware.ts
@@ -1,4 +1,4 @@
-import { api, utils } from '@hoprnet/hopr-sdk';
+import { utils } from '@hoprnet/hopr-sdk';
 import { Middleware, PayloadAction } from '@reduxjs/toolkit';
 import { initialState } from '../auth/initialState';
 import { nodeActions } from './index';
@@ -10,7 +10,6 @@ const {
   updateWebsocketStatus,
 } = nodeActions;
 
-const { getWsUrl } = api;
 const { WebsocketHelper } = utils;
 
 type LocalRootState = {
@@ -28,14 +27,10 @@ const websocketMiddleware: Middleware<{}, LocalRootState> = ({
       // start websocket connection
       const { apiEndpoint, apiToken } = getState().auth.loginData;
       if (apiEndpoint && apiToken) {
-        const wsUrl = getWsUrl(
-          apiEndpoint,
-          '/api/v2/messages/websocket/',
-          apiToken
-        );
         try {
           socket = new WebsocketHelper({
-            url: wsUrl,
+            apiEndpoint,
+            apiToken,
             onOpen: () => {
               dispatch(updateWebsocketStatus(true));
             },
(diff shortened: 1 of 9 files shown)


diegoalzate:
blocked by https://github.com/hoprnet/hopr-sdk/pull/56

mjadach-iv:
Lets wait with merge until older PRs are merged.

mjadach-iv (changes requested):
Rebase needed

src/store/slices/node/websocketMiddleware.ts
```diff
@@ -0,0 +1,70 @@
...
+  let socket: typeof WebsocketHelper.prototype | null = null;
+
+  return (next) => (action: PayloadAction) => {
+    if (action.type === initializeWebsocket.type) {
+      // start websocket connection
+      const { apiEndpoint, apiToken } = getState().auth.loginData;
+      if (apiEndpoint && apiToken) {
+        const wsUrl = getWsUrl(
+          apiEndpoint,
+          '/api/v2/messages/websocket/',
```
mjadach-iv, line 33: shouldn't this part be provided by the HOPR-SDK?

src/store/slices/node/websocketMiddleware.ts
```diff
@@ -0,0 +1,70 @@
...
+          apiToken
+        );
+        try {
+          socket = new WebsocketHelper({
+            url: wsUrl,
```
mjadach-iv, line 38: We should update the WebsocketHelper to comly with the apiEndpoint and apiToken standard we have in all the other calls, and then use it here

mjadach-iv (approved):

PR #27, opened 2023-06-01
Aliases subpage

### LINKED ISSUE #16: Admin subpage: Aliases
Definition of done:
 - no styling, plain html
 - table with 4 columns: id (from 0 up), peerId, alias, actions
 - import and export (into csv file) buttons
 - add alias button
 - remove alias button (in actions of the table)

diegoalzate:
closes #16 

### overview
provides aliases interaction like deleting/creating aliases. this pr contains no specific styles, the table is created with mui so it definitely looks better than the rest of things.

#### things to know moving forward
- created a component to import csv data that can be re used
- created a export to csv function that can also be re used

### DIFF
--- a/src/sections/aliases.tsx
+++ b/src/sections/aliases.tsx
@@ -1,0 +1,211 @@
+import { Paper, Table, TableBody, TableCell, TableContainer, TableHead, TableRow } from "@mui/material";
+import Section from "../future-hopr-lib-components/Section";
+import { useAppDispatch, useAppSelector } from "../store";
+import { useEffect, useRef, useState } from "react";
+import { actionsAsync } from "../store/slices/node/actionsAsync";
+import { exportToCsv } from "../utils/helpers";
+
+function AliasesPage() {
+  const dispatch = useAppDispatch();
+  const aliases = useAppSelector(selector => selector.sdk.aliases)
+  const loginData = useAppSelector(selector => selector.auth.loginData)
+
+  useEffect(() => {
+    if (loginData.ip && loginData.apiKey) {
+      dispatch(actionsAsync.getAliasesThunk({ apiEndpoint: loginData.ip, apiToken: loginData.apiKey }))
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [loginData])
+
+  return (
+    <Section
+      className="Section--aliases"
+      id="Section--aliases"
+      yellow
+    >
+      <h2>Add new alias</h2>
+      <CreateAliasForm />
+      <h2>Aliases table</h2>
+      <button onClick={() => {
+        if (aliases) {
+          exportToCsv(Object.keys(aliases).map(alias => ({
+            alias: alias,
+            peerId: aliases[alias]
+          })), "aliases.csv")
+        }
+      }}>export</button>
+      <CSVUploader onParse={(parsedData) => {
+        for (const data of parsedData) {
+          if (data.alias && data.peerId && loginData.ip && loginData.apiKey) {
+            dispatch(actionsAsync.setAliasThunk({
+              alias: String(data.alias),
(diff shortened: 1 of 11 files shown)


src/sections/aliases.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+    // Reset the file input value
+    if (fileInputRef.current) {
+      fileInputRef.current.value = "";
+    }
+  };
+
+  return (
+    <div>
+      import <input type="file" accept=".csv" ref={fileInputRef} onChange={handleFileUpload} placeholder="import" />
```
mjadach-iv, line 205: Here just add an 'Import' button which will open the File Explorer for importing.
All that without the import text and 'no file chosen' etc.

src/sections/aliases.tsx
```diff
@@ -0,0 +1,211 @@
...
+    setForm({ ...form, [name]: value })
+  }
+
+  return (
+    <div>
+      <input type="text" name="peerId" placeholder="peerId" onChange={handleChange} value={form.peerId} />
+      <input type="text" name="alias" placeholder="alias" onChange={handleChange} value={form.alias} />
+      <button onClick={() => {
+        if (loginData.ip && loginData.apiKey) {
+          dispatch(actionsAsync.setAliasThunk({ alias: form.alias, peerId: form.peerId, apiEndpoint: loginData.ip, apiToken: loginData.apiKey }))
```
mjadach-iv, line 110: If that call will fail, we need an error to show up under the form with information what went wrong.
If the call will succeed, we show information under the form that it;'s saved and we clear the textfield.

src/sections/aliases.tsx
```diff
@@ -0,0 +1,211 @@
...
+            {Object.entries(aliases ?? {}).map(([alias, peerId], key) => (
+              <TableRow
+                key={key}
+              >
+                <TableCell component="th" scope="row">
+                  {key}
+                </TableCell>
+                <TableCell>{peerId}</TableCell>
+                <TableCell>{alias}</TableCell>
+                <TableCell><DeleteAliasButton alias={alias} /></TableCell>
```
mjadach-iv, line 69: We need a state in the component in which we will store 'table action errors'.
If this call fails, save the message to this state.
In next PRs we will add Notification center or information snackbar center.

src/sections/aliases.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+  return (
+    <Section
+      className="Section--aliases"
+      id="Section--aliases"
+      yellow
+    >
+      <h2>Add new alias</h2>
+      <CreateAliasForm />
+      <h2>Aliases table</h2>
```
mjadach-iv, line 28: in this line, please also add a refresh button, which will download aliases again

mjadach-iv (changes requested):

src/sections/aliases.tsx
diegoalzate: I can do this by hidding the <input /> and triggering it through a button but not sure how a button can open the file explorer without the <input />

ItsLuized (approved):

src/sections/aliases.tsx
```diff
@@ -0,0 +1,344 @@
...
+      );
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [loginData]);
+
+  return (
+    <Section className="Section--aliases" id="Section--aliases" yellow>
+      <h2>Add new alias</h2>
+      <CreateAliasForm />
+      <h2>Aliases table</h2>
```
mjadach-iv, line 42: Under the table please add the same error handling as you added in the Add new alias

mjadach-iv (changes requested):

mjadach-iv (approved):

PR #39, opened 2023-06-05
Logs subpage

camilovegag:
### Description:

Logs subpage with info coming from a web socket with path `'/api/v2/node/stream/websocket/'` created at `websocketMiddleware`.

It uses `readStreamEvent` to convert the web socket messages to client logs.

### Evidence:
image


camilovegag:
Missing new SDK version and some merge conflicts

Current state:
image

src/components/JazzIcon/JazzIcon.tsx
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+export type Props = {
+  diameter: number | null | undefined;
+  className?: string;
+  style?: Partial<CSSProperties>;
+  address?: string;
+};
+
+export default class Jazzicon extends PureComponent<Props> {
```
diegoalzate, line 22: is there a reason this is a class component?

camilovegag: Updated Jazzicon to a functional component

diegoalzate (approved):

camilovegag:
Need to update the logic for abbr ids, so it works on all types of logs and not only

`Node started: 16Uu....`

src/components/AbbreviatedId/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+  font-weight: bold;
+`;
+
+const AbbrLogId = LogId.withComponent('abbr');
+
+type AbbreviatedIdProps = {
+  id: string;
+};
+
+const AbbreviatedId = ({ id }: AbbreviatedIdProps) => {
```
mjadach-iv, line 15:
```suggestion
const AbbreviatedPeerId = ({ id }: AbbreviatedIdProps) => {
```

src/components/JazzIcon/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+    const icon = jazzicon(diameter, md5(address));
+    containerRef.current?.appendChild(icon);
+    return () => {
+      if (containerRef.current) {
+        containerRef.current.innerHTML = '';
+      }
+    };
+  }, []);
+
+  return <div ref={containerRef} />;
```
mjadach-iv, line 24:
```suggestion
  return (
    <div
      className={`Jazzicon ${className}`}
      ref={containerRef} 
      {...rest}
    />
  );
```

src/components/JazzIcon/index.tsx
```diff
@@ -0,0 +1,27 @@
+import { useEffect, useRef } from 'react';
+import md5 from 'md5';
+// @ts-ignore
+import jazzicon from '@metamask/jazzicon';
```
mjadach-iv, line 4: It's a good idea for a component. Lets add it to the [future-hopr-lib-components](https://github.com/hoprnet/hopr-admin/tree/develop/src/future-hopr-lib-components) folder

src/router.tsx
```diff
@@ -31,6 +32,12 @@ export const applicationMap = [
         icon: <MenuIcon />,
         element: <Section1 />,
       },
+      {
+        name: 'Logs',
+        path: 'logs',
+        icon: <MailIcon />,
+        element: <SectionLogs />,
```
mjadach-iv, line 39: Please rebase the PR and add the loginNeeded value here

src/components/LogLine/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+            <div>
+              {log.message.replace(peerId, '')}
+              <AbbreviatedId id={peerId} />
+            </div>
+          </>
+        ) : (
+          log.message
+        )}
+      </Pre>
+      <Time>{timestamp}</Time>
```
mjadach-iv, line 55: - lets make it a component
- place the component for now inside this file
- wrap the <Time> with mui tooltip
- provice timestamp as a props to the <Time> componrnt
- print the value like you do it now: `{timestamp}` with  `const timestamp = new Date(log.timestamp).toISOString().slice(11);`
- in the tooltip provide a full string of the date: e.g `Mon Jun 12 2023 11:18:47 GMT+0200 (Central European Summer Time)` that you can get by eq. of `new Date():`

mjadach-iv (changes requested):

src/components/JazzIcon/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+type JazziconProps = {
+  address: string;
+  diameter: number;
+};
```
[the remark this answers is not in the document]
camilovegag: I'll be doing `interface JazzIconProps extends HTMLAttributes<HTMLDivElement>`, I readed about `React.FC<JazzIconProps>` but it was considered an anti-pattern because it allowed components to accept children even when they weren't supposed to.

Seems that the FC hate has disappeared starting from React 18 and TS 5.1 as it children is no longer included in the props type, I think we can discuss this on tomorrows daily :)

camilovegag:
Updated with review

mjadach-iv (approved):

PR #44, opened 2023-06-07
add safe tx redux actions

### LINKED ISSUE #43: add Safe transaction and setting actions to redux
definition of done:
- [ ] create a tx proposal
- [ ] accept a tx proposal
- [ ] get all txs
- [ ] get safe information

diegoalzate:
closes #43 

### overview
adds redux actions to create, reject and get all transactions. Also get all the safe information.

this pr includes some ui buttons and actions but that should be handled again in another pr


src/sections/safe.tsx
```diff
@@ -8,21 +7,117 @@ import { useAppDispatch, useAppSelector } from '../store';
 import Section from '../future-hopr-lib-components/Section';
 import { useSigner } from '../hooks';
 import { actionsAsync } from '../store/slices/safe/actionsAsync';
```
camilovegag, line 9: Move this under // Stores

src/sections/safe.tsx
```diff
@@ -8,21 +7,117 @@ import { useAppDispatch, useAppSelector } from '../store';
...
   useEffect(() => {
     if (signer) {
       dispatch(actionsAsync.getSafesByOwnerThunk({ signer }));
     }
   }, [signer]);
+
   return (
     <Section className="Section--safe" id="Section--safe" yellow>
-      SAFE REDUX STORE
+      <h1>SAFE REDUX STORE</h1>
```
camilovegag, line 24: Just a comment, when we work on the UI we might want to handle uppercase text with styles `text-transform: uppercase;`  :)

src/sections/safe.tsx
diegoalzate: i can change this in this pr but it was not part of the existing changes.

src/sections/safe.tsx
diegoalzate: yeah the uppercase was also not part of this pr

src/sections/safe.tsx
```diff
@@ -1,28 +1,123 @@
...
     if (signer) {
       dispatch(actionsAsync.getSafesByOwnerThunk({ signer }));
     }
   }, [signer]);
+
   return (
     <Section className="Section--safe" id="Section--safe" yellow>
-      SAFE REDUX STORE
+      <h1>Safe redux store</h1>
       <pre>{JSON.stringify(safe, null, 4)}</pre>
```
mjadach-iv, line 25: Please switch around the store and actions, so `<pre>{JSON.stringify(safe, null, 4)}</pre>` is below actions

mjadach-iv (changes requested):
If a wallet is not connected in the UI, the `develop/safe` shouldn't work

mjadach-iv (approved):

PR #40, opened 2023-06-06
Add tickets page with redeemTickets function

ItsLuized:
Closes #23

### Overview

This pull Request provides tickets interaction like redeemingTickets, getting unredeemed tickets and redeemed tickets.

Things to be added:
- [x] Displays information about your redeemed and unredeemed tickets
- [x] Button: Redeems all your tickets
- [ ] Add buttons for all sdk tickets functions
- [x] Refresh button needed at the same line as title of the page, button should be on the right

PR #45, opened 2023-06-08
Peers subpage

### LINKED ISSUE #21: Admin subpage: Peers
Definition of done:
- no styling, plain html
- title: `Peers seen in the network (${peers.length})`
- table with columns: id (from 0 up), peerId, alias, Eligible, Multiaddrs

ItsLuized:
Closes #21

### Overview
This pull Request provides peers subpage to show the announces peers with their info

Things to be added:

- [x] no styling, plain html
- [x] title: Peers seen in the network (${peers.length})
- [x] table with columns: id (from 0 up), peerId, alias, Eligible, Multiaddrs
- [x] refresh button needed at the same line as title of the page, button should be on the right

To be noted:

Elegible is currently set to show true if the peerId is connected to the current node (in both the announced and connected arrays of peers)

### Current state
image

### DIFF
--- a/src/sections/peers.tsx
+++ b/src/sections/peers.tsx
@@ -66,7 +66,7 @@ function PeersPage() {
           </TableHead>
           <TableBody>
             {Object.entries(peers?.announced ?? {}).map(([id, peer], key) => (
-              <TableRow key={key}>
+              <TableRow key={id}>
                 <TableCell component="th" scope="row">
                   {id}
                 </TableCell>
--- a/src/sections/peers.tsx
+++ b/src/sections/peers.tsx
@@ -1,0 +1,109 @@
+import {
+  Paper,
+  Table,
+  TableBody,
+  TableCell,
+  TableContainer,
+  TableHead,
+  TableRow,
+} from '@mui/material';
+import { useEffect } from 'react';
+import Section from '../future-hopr-lib-components/Section';
+import { useAppDispatch, useAppSelector } from '../store';
+import { actionsAsync } from '../store/slices/node/actionsAsync';
+
+function PeersPage() {
+  const dispatch = useAppDispatch();
+  const loginData = useAppSelector((selector) => selector.auth.loginData);
+  const peers = useAppSelector((selector) => selector.sdk.peers);
+  const aliases = useAppSelector((selector) => selector.sdk.aliases);
+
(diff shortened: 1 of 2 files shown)


src/sections/peers.tsx
```diff
@@ -0,0 +1,109 @@
...
+        apiEndpoint: loginData.apiEndpoint!,
+        apiToken: loginData.apiToken!,
+      })
+    );
+    dispatch(
+      actionsAsync.getAliasesThunk({
+        apiEndpoint: loginData.apiEndpoint!,
+        apiToken: loginData.apiToken!,
+      })
+    );
```
camilovegag:
```ts
if (loginData) {
   handleRefresh();
}
```

src/sections/peers.tsx
```diff
@@ -0,0 +1,109 @@
...
+  };
+
+  return (
+    <Section>
+      <h2>
+        Peers seen in the network ({peers?.announced?.length || 0}){' '}
+        <button
+          onClick={() => {
+            handleRefresh();
+          }}
```
camilovegag, line 56: This can be changed just to :
```ts
onClick={handleRefresh}
```

src/sections/peers.tsx
```diff
@@ -0,0 +1,109 @@
...
+              <TableCell>id</TableCell>
+              <TableCell>peerId</TableCell>
+              <TableCell>alias</TableCell>
+              <TableCell>Elegible</TableCell>
+              <TableCell>Multiaddrs</TableCell>
+            </TableRow>
+          </TableHead>
+          <TableBody>
+            {Object.entries(peers?.announced ?? {}).map(([id, peer], key) => (
+              <TableRow key={key}>
```
camilovegag, line 75: Using `id` as the `key` here might be better than using the second map argument `key` if we change the order of the array in the future

camilovegag (changes requested):

src/sections/peers.tsx
ItsLuized: Regarding this, in a different PR, this was implemented in a similar way and in the Review it was adviced to remove it:
https://github.com/hoprnet/hopr-admin/pull/40#discussion_r1221068542

camilovegag (approved):

PR #46, opened 2023-06-08
add eslint configuration

diegoalzate:
### overview
with the migration to vite, eslint was forgotten, this pr adds eslint back to the project.

### what does eslint do and not do?
not: format code and force a code style
does: checks syntax and finds possible problems

### DIFF
--- a/src/sections/info.tsx
+++ b/src/sections/info.tsx
@@ -21,7 +21,6 @@ function InfoPage() {
 
   useEffect(() => {
     fetchInfoData();
-    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
   const fetchInfoData = () => {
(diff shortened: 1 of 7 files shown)


src/sections/info.tsx
```diff
@@ -54,21 +53,21 @@ function InfoPage() {
       <div id="info">
         <div id="version">Version: {version}</div>
         {!!info &&
-          Object.entries(info).map(([key, value]) => {
+          Object.entries(info).map(([key, value], index) => {
```
camilovegag, line 56: Using `index` as`key` might not be good if the order of items may change, can this happen here?

ItsLuized (approved):
Nice catch, eslint always good to have! 👍🏻

src/sections/info.tsx
camilovegag: gotcha ;)

camilovegag (approved):

PR #47, opened 2023-06-09
Messages subpage

camilovegag:


camilovegag:
Working with aliases now, missing seen status as:

`Cannot assign to read only property 'seen' of object '#<Object>'`

mjadach-iv (approved):

PR #55, opened 2023-06-14
add pre commit changes to commit

diegoalzate:
### overview
this pr adds pre-commit changes to current commit.

before commits would look like this:
1. added nice feature
2. format

after:
1. added nice feature with format changes


diegoalzate:
stale

PR #56, opened 2023-06-15
Prettier + ESlint plugins

camilovegag:
### Description: 
Adding Prettier options and ESlint plugins and rules in order to have single line imports and object destructuring.

### Evidence:
- imports: Screenshot 2023-06-14 at 8 41 39 PM

- destructuring: Screenshot 2023-06-14 at 8 46 18 PM

## For better vs code results, use Prettier ESLint extension that formats code on save taking the best from both worlds.

We might find weird behaviors as we use this new setup, tried to find and fix the most commons but feedback will def help us out to better improve our code rules

Adding eslint rule for console.logs too in this pr, not error, just warning :)


.eslintrc.json
```diff
@@ -16,9 +12,30 @@
...
+    "newline-destructuring/newline": "warn",
+    "no-duplicate-imports": "warn",
+    "no-console": "warn",
+    "prefer-const": "warn",
+    "no-var": "off"
+  },
+  "settings": {
+    "react": {
+      "version": "detect"
+    }
```
diegoalzate, line 39: nice work @camilovegag.

#### this is a very subjective opinion:

I think eslint should not care about formatting. Eslint should check for code quality and prettier should take care of this. However, i know that the plugins that were added to this config are not available as prettier config, so i suggest we scrap that from the requirement that we have for formatting.

my ideal settings for eslint:
```json
{
  "env": {
    "browser": true,
    "es2021": true
  },
  "extends": ["eslint:recommended", "plugin:react/recommended", "plugin:@typescript-eslint/recommended"],
  "parser": "@typescript-eslint/parser",
  "parserOptions": {
    "ecmaFeatures": {
      "jsx": true
    },
    "ecmaVersion": "latest",
    "sourceType": "module"
  },
  "plugins": ["react", "@typescript-eslint", "prettier"],
  "rules": {
    "react/react-in-jsx-scope": "off",
    "react/jsx-filename-extension": [1, { "extensions": [".ts", ".tsx"] }],
    "no-console": "warn",
    "no-duplicate-imports": "warn",
    "prefer-const": "warn"
  }  
}

```

cc @mjadach-iv

camilovegag:
Why does it says:

```
All checks have failed
2 failing checks

fleek/build — Build failed
hopr-admin Failing after 2m — Build failed
```
🤔

mjadach-iv:
The repo has connected checks because the main branch is a real main for the Admin UI which is currently in use.

PR #57, opened 2023-06-15
Admin channels subpage

### LINKED ISSUE #17: Admin subpage: Channels
Definition of done:
- no styling, plain html
- 2 tabs incomming/outgoing with #tab-name added to url (use mui tabs)
- on each tab tables with columns: id (from 0 up), peerId, status, dedicated funds, actions
- export (into csv file) buttons on each tab
- aliases displayed in place of peersid (with peer id added as alt attribute in html) in the tables, but not in the export file
- `open channel` button over outgoing channels table
- close channel button in actions in table
- fund channel button in actions in table

ItsLuized:
Closes #17 

### Overview

This pull request adds the admin subpage for channels and its functionality

This to be added:

- [x] no styling, plain html
- [x] 2 tabs incomming/outgoing with #tab-name added to url (use mui tabs)
- [x] on each tab tables with columns: id (from 0 up), peerId, status, dedicated funds, actions
- [x] export (into csv file) buttons on each tab
- [x] aliases displayed in place of peersid (with peer id added as alt attribute in html) in the tables, but not in the export file
- [x] open channel button over outgoing channels table
- [x] close channel button in actions in table
- [x] fund channel button in actions in table
- [x] refresh button needed at the same line as title of the page, button should be on the right


src/router.tsx
```diff
@@ -78,6 +79,12 @@ export const applicationMap = [
         element: <TicketsPage />,
         loginNeeded: 'node',
       },
+      {
+        name: 'Channels',
+        path: 'channels',
+        icon: <MailIcon />,
+        element: <ChannelsPage />,
+      },
```
mjadach-iv, line 87: Below you have a Channels already

src/sections/channels.tsx
```diff
@@ -0,0 +1,492 @@
...
+            <TableBody>
+              {Object.entries(channels?.outgoing ?? []).map(
+                ([, channel], key) => (
+                  <TableRow key={key}>
+                    <TableCell component="th" scope="row">
+                      {key}
+                    </TableCell>
+                    <TableCell>{getAliasByPeerId(channel.peerId)}</TableCell>
+                    <TableCell>{channel.status}</TableCell>
+                    <TableCell>{channel.balance}</TableCell>
```
mjadach-iv, line 454: We need a function in utils that will format balances. Something like format `formatEther(balance)`, as we do not have `500000000000000000` but  `0.5 mHOPR`. Also add `mHOPR` at the end here.

src/sections/channels.tsx
```diff
@@ -0,0 +1,492 @@
...
+                error: e.error,
+                status: e.status,
+              },
+            ],
+          },
+        }));
+      });
+  };
+
+  const fundChannel = (channelId: string, peerId: string) => {
```
mjadach-iv, line 289: if its a React Component, it should start with a big letter, and also it should be more clear what it does like: `FundChannelModal`.
Put it into the components folder, as we will you it in more sections

src/sections/channels.tsx
```diff
@@ -0,0 +1,492 @@
...
+    return (
+      <>
+        <button onClick={() => handleOpenFundingPopup(channelId)}>Fund</button>
+        <Dialog
+          open={openFundingPopups[channelId] || false}
+          onClose={() => closeFundingPopup(channelId)}
+        >
+          <DialogTitle>Fund Channel</DialogTitle>
+          <DialogContent>
+            <TextField
```
mjadach-iv, line 299: We need a `mHOPR` currency here. The Mui TextField has this option as props.
Also we should parse the input here, as when we add '1', what we mean is 1 mHOPR and not 0.000000000000000001 mHOPR

src/sections/channels.tsx
```diff
@@ -0,0 +1,492 @@
...
+      <h2>
+        Channels
+        <button onClick={handleRefresh}>Refresh</button>
+      </h2>
+      <Box sx={{ borderColor: 'divider' }}>
+        <Tabs value={tabIndex} onChange={handleTabChange} centered>
+          <Tab label="Incoming" />
+          <Tab label="Outgoing" />
+        </Tabs>
+      </Box>
```
mjadach-iv, line 377: You need to handle the params in the url here. If we go on the Channels subpage, you lose the login credentials from the omnibar

src/sections/channels.tsx
```diff
@@ -0,0 +1,492 @@
...
+            />
+          </DialogContent>
+          <DialogActions>
+            <button onClick={() => closeFundingPopup(channelId)}>Cancel</button>
+            <button
+              onClick={() => handleFundChannels(peerId, channelId)}
+              disabled={!fundingAmount || parseFloat(fundingAmount) <= 0}
+            >
+              Fund
+            </button>
```
mjadach-iv, line 313: After clicking, you need to disable the button and show loader, so we know it was clicked or just close the popup right away. Lets go with the 2nd option.

mjadach-iv (changes requested):

mjadach-iv:
Is there a function for opening multiple channels? If there is, we also need an import button that will trigger it

ItsLuized:
> Is there a function for opening multiple channels? If there is, we also need an import button that will trigger it

There is currently not an open multiple channels functionality. This will be added on hoprnet/hopr-sdk#38.

I will create a issue that has this Pull request as a blocker so we can tackle this when it's merged.

mjadach-iv (approved):

PR #60, opened 2023-06-15
Add watcher

### LINKED ISSUE #24: Admin: Watcher
Definition of Done:
 - channel is funded
 - channel is closed
 - node is funded
 - netowork state changed
 - when message is received
 - redux slice for notifications (array?)
 - there might be other thigs to watch for (WIP)

diegoalzate:
closes #24 

### overview
Create watcher component that compares previous states of certain variables with new updates and dispatches notifications on updates.

#### how?
Using a temporal variable we check every certain amount of time if the new variable is different to the last saved temporal variable.  

### edge case
messages, because there is no way to fetch an update we just have to wait for the redux store to update with some new messages.

### DIFF
--- a/src/providers/Watcher/index.tsx
+++ b/src/providers/Watcher/index.tsx
@@ -211,8 +211,6 @@ const Watcher = () => {
   };
 
   const watchMessages = () => {
-    console.log({ prevMessages }, { messages });
-
     if (prevMessages && prevMessages.length < messages.length) {
       dispatch(
         appActions.addNotification({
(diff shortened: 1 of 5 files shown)


src/providers/Watcher/index.tsx
```diff
@@ -0,0 +1,310 @@
...
+  // booleans to stop interval from running in parallel
+  let watchChannelsIntervalIsRunning = false;
+  let watchNodeFundsIntervalIsRunning = false;
+  let watchNodeInfoIntervalIsRunning = false;
+
+  useEffect(() => {
+    // reset state on every change of node
+    if (
+      prevLoginData?.apiEndpoint !== apiEndpoint ||
+      prevLoginData.apiToken !== apiToken
```
mjadach-iv, line 46: That might backfire if someone will change apiToken for of the node and re-log.
Since we only work on notifications on the Front-End side, let's leave it for now. Maybe we will get BE based feature like that

mjadach-iv (approved):

PR #69, opened 2023-06-20
Fix loggin out credential bug and the # in the url error

### LINKED ISSUE #63: Channels subpage: #incoming and #outgoing in url do not work
Fix the follwoing:
 - When copying the url from 1 tab and going onto another, you should be placed in the same part of the application. Now you are always placed on the #incoming tab. 
 - when logging off on the channels subpage, we get logged in right away with `null` credentials

ItsLuized:
Closes #63

### Overview

This pull request aims to fix the bugs on the admin subpage: Channels.
Fixes expected:

- [x] When copying the url from 1 tab and going onto another, you should be placed in the same part of the application. Now you are always placed on the #incoming tab.
- [x] When logging off on the channels subpage, we get logged in right away with null credentials

### DIFF
--- a/src/sections/channels.tsx
+++ b/src/sections/channels.tsx
@@ -157,20 +157,28 @@ function ChannelsPage() {
   };
 
   useEffect(() => {
-    const queryParams = new URLSearchParams({
-      apiToken: loginData.apiToken!,
-      apiEndpoint: loginData.apiEndpoint!,
-    }).toString();
-
-    set_queryParams(queryParams);
-
-    if (queryParams) {
-      console.log(`queryParams: ${queryParams}`);
-      navigate(`?${queryParams}#incoming`, { replace: true });
+    if (loginData.apiEndpoint && loginData.apiToken) {
+      const queryParams = new URLSearchParams({
+        apiToken: loginData.apiToken,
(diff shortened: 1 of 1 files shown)


src/sections/channels.tsx
```diff
@@ -157,20 +157,28 @@ function ChannelsPage() {
...
-      console.log(`queryParams: ${queryParams}`);
-      navigate(`?${queryParams}#incoming`, { replace: true });
-    }
+    const defaultTabIndex = defaultHash === '#outgoing' ? 1 : 0;
+    setTabIndex(defaultTabIndex);
+    handleHash(defaultTabIndex);
 
     handleRefresh();
-  }, [loginData.apiToken, loginData.apiEndpoint, navigate]);
+  }, [queryParams]);
```
camilovegag, line 181: is this dependency needed here?

ItsLuized: Yes because we need to make sure the queryParams are set before executing `handleHash` which adds the queryParams and hash to the url.

If the dependency is not there, the first time we enter the channels subpage we would get the url without the credentials

mjadach-iv (approved):

PR #73, opened 2023-06-21
Bugfix/node login

mjadach-iv:
Fix for:
- the select node didn't work before. When clicking on it, it didn't change.
- the `Save API Key locally (unsafe)` did mark itself when the secret wasn't saved locally.
- the `use node` button didn't use Redux fullfiled/rejected logic
- the url with apiEndpoint but without apiToken did not populate textfields

### DIFF
--- a/src/sections/selectNode.tsx
+++ b/src/sections/selectNode.tsx
@@ -57,7 +57,8 @@ function Section1() {
     const existingItemIndex = nodesSavedLocally.findIndex((item: any) => item.apiEndpoint === loginData.apiEndpoint);
     if (existingItemIndex !== -1) set_nodesSavedLocallyChosenIndex(existingItemIndex.toString());
     const existingItem = nodesSavedLocally[existingItemIndex] as ParsedNode;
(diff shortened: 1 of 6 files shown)


src/sections/selectNode.tsx
```diff
@@ -13,35 +13,52 @@ import Checkbox from '../future-hopr-lib-components/Toggles/Checkbox';
...
   const dispatch = useAppDispatch();
-  const nodesSavedLocally = useAppSelector((store) => store.auth.nodes).map((node, index) => {
-    return {
-      name: node.localName ? `${node.localName} (${node.apiEndpoint})` : node.apiEndpoint,
-      localName: node.localName,
-      value: index,
-      apiEndpoint: node.apiEndpoint,
-      apiToken: node.apiToken,
-    };
-  });
```
mjadach-iv, line 27: It was triggering useEffect with [nodesSavedLocally] constantly

ItsLuized (approved):

PR #77, opened 2023-06-26
Redux fixes for loging in and changing nodes

mjadach-iv:


diegoalzate:
Is there an issue for this?

diegoalzate (approved):
I see that we know will close ws and reset state when disconnecting from a node. looks good at this point.

mjadach-iv:
There was no issue. Thx for the review

PR #82, opened 2023-06-28
Add toasts

### LINKED ISSUE #48: Admin UI: Toasts
Definition of done:
 - use of Toastify: https://fkhadra.github.io/react-toastify/use-react-toastify-with-redux
 - similar to that: https://codesandbox.io/s/2485wxy78j?from-embed
 - when a new notification is added to Redux, show the toast at the bottom right
 - if multiple notifications come, add new one at the bottom right and move the last one up by one
 - 5s notification time
 - support for all types: info, success, warning, error, default
 - notifications should have 'seen' (instead of read) and 'new' keys in the object
 - if 'x' is clicked on the toast, then 'seen' and 'new' should be true
 - if a notification icon is pressed on the app bar, all the notifications should have be mared as 'new: false'
 - notification that is not 'seen' should have a background highlight of hopr blue with some transparency
 - if you click on the notification in the list in the app bar, it shoud be marked as 'seen: true'
 - the number on the notification bell in the app bar should count the 'new: true' notifications
- add more details on the toast than we have at the appbar dropdown

diegoalzate:
closes #48 

### overview
add toasts that work together with notifications.

#### changes in redux
```
{
  seen: bool // is true until notification tray is opened, when notification tray is opened all notifications will be seen but not read
  read: bool // true when notification is clicked
}
```

changes to ux:
- when a user clicks on a toast it is marked as read and seen
- when a user opens notification tray all notifications are marked as seen
- notifications that are marked as **not** read will have a blue background


src/App.tsx
```diff
@@ -1,24 +1,18 @@
...
 import Watcher from './providers/Watcher';
+import { ToastContainer } from 'react-toastify';
+import 'react-toastify/dist/ReactToastify.css';
 
 function App() {
   return (
     <Provider store={store}>
       <WagmiProvider>
         <Watcher />
+        <ToastContainer position="bottom-right" />
```
camilovegag, line 15: Maybe adding this to our Layout can be better? just guessing

>Remember to render the ToastContainer once in your application tree. If you can't figure out where to put it, rendering it in the application root would be the best bet.

src/providers/Watcher/index.tsx
```diff
@@ -64,6 +67,31 @@ const Watcher = () => {
...
 
+  const sendNotification = ({
+    notificationPayload,
+    toastPayload,
+  }: {
+    // notification payload without id
+    notificationPayload: Omit<Parameters<typeof appActions.addNotification>[0], 'id'>;
+    toastPayload?: { message: string; type?: ToastOptions['type'] };
+  }) => {
+    const notificationId = nanoid();
```
camilovegag, line 78: Do we have a redux id already? When you were off @mjadach-iv talked about `nanoid` but I can't quite remember it

src/providers/Watcher/index.tsx
```diff
@@ -155,14 +187,15 @@ const Watcher = () => {
...
+            notificationPayload: {
               source: 'node',
               name: notificationText,
-              timeout: null,
               url: null,
-            }),
-          );
+              timeout: null,
+            },
+            toastPayload: { message: updatedChannel.channelId + ' ' + notificationText },
```
camilovegag, line 197: You used template literals above, I think we can do it here too :)

src/providers/Watcher/index.tsx
```diff
@@ -172,21 +205,32 @@ const Watcher = () => {
...
+        notificationPayload: {
           source: 'node/message',
           name: 'Received new message',
-          timeout: null,
           url: null,
-        }),
-      );
+          timeout: null,
+        },
+        toastPayload: { message: 'received message: ' + newMessage.latestMessage.body },
```
camilovegag, line 232: same

vite.config.ts
```diff
@@ -5,7 +5,7 @@ import svgrPlugin from 'vite-plugin-svgr';
 import { NodeGlobalsPolyfillPlugin } from '@esbuild-plugins/node-globals-polyfill';
 
 // https://vitejs.dev/config/
-export default defineConfig(({ mode }) => {
+export default defineConfig(() => {
```
camilovegag, line 8: `mode` removed by formatter?

src/App.tsx
diegoalzate: In this case I think it is our decision. App should only be rendered once in the app so it would be fine if that is the requirement

src/providers/Watcher/index.tsx
diegoalzate: had to move it out of redux because the notification toast also needed to get the id to know what notification was clicked. That is why it is required in the payload now.

vite.config.ts
diegoalzate: i removed it

camilovegag:
### RPCh node
Connected with an RPCh node and got bombarded with toasts, but they are over our appbar, is possible to add some padding so they don't overlap it?

PD: Right drawer totally covered lol

Screenshot 2023-06-29 at 6 03 37 PM

### Height

Also I think the notification tray is to big (27" monitor), adding some max height can fix it.

Screenshot 2023-06-29 at 6 05 49 PM

### Crash

This might be due to the giant amount of messages this node gets, but some toasts freeze (my app crashed), thing to keep in mind in the future 

Screenshot 2023-06-29 at 6 09 14 PM

src/App.tsx
mjadach-iv: Looks good for me here

src/components/NotificationBar/index.tsx
```diff
@@ -81,22 +88,34 @@ export default function NotificationBar() {
...
+                  dispatch(appActions.interactedWithNotification(notification.id));
+                  if (notification.source === 'node/message') {
+                    navigate(`networking/messages${searchParams ? searchParams : ''}`);
+                  }
+                }}
+              >
+                {notification.name}
+              </StyledUnreadMenuItem>
+            ),
+          )
```
mjadach-iv, line 118: Those 2 are the same. Just manipulate the styling with a className like: ```className={`notification.interacted ? 'unreadMenuItem' : ''`}``` and add in styled sth like:
```
const SMenuItem = styled(MenuItem)`
&.unreadMenuItem {
  background-color: rgba(0, 0, 180, 0.3);
  opacity: 90%;
  border-bottom: 1px solid #444;
  }
`;
```

src/providers/Watcher/index.tsx
mjadach-iv: You get id from redux toolkit on every action, check the messages async actions. If you can't you the same logic for getting id, we can stay with this one.

mjadach-iv (changes requested):

src/providers/Watcher/index.tsx
mjadach-iv: It's a good implementation

mjadach-iv (approved):

Issue #34, opened 2023-06-05
Admin wrapper: new subpage


mjadach-iv:
Definition of done:
- add Wrapper subpage to 'DEVELOP' section in the drawer
- mirror the wrapping functonality of the https://wrapper.hoprnet.org/ by writing new code (not a copy/paste) on the new Wrapper subpage

UI:
https://www.figma.com/proto/qjTbRGulmsPBQ2E0aHY7HI/Staking-UI-v3?type=design&node-id=647-6493&scaling=min-zoom&page-id=0%3A1&starting-point-node-id=553%3A3600&show-proto-sidebar=1

PR #83, opened 2023-06-29
Subpage: Wrapper

camilovegag:
### Closes: #34

### DIFF
--- a/src/sections/wrapper.tsx
+++ b/src/sections/wrapper.tsx
@@ -1,0 +1,86 @@
+import { IconButton, Paper, TextField } from '@mui/material';
+import Section from '../future-hopr-lib-components/Section';
+import styled from '@emotion/styled';
+import Button from '../future-hopr-lib-components/Button';
+
+import SwapVertIcon from '@mui/icons-material/SwapVert';
+import { useState } from 'react';
+
+const StyledPaper = styled(Paper)`
+  &.MuiPaper-root {
+    padding: 2rem;
+    text-align: center;
+  }
+`;
+
+const FlexContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 1rem;
+  position: relative;
+
+  & button {
+    align-self: center;
+    text-transform: uppercase;
+  }
+`;
+
+const StyledIconButton = styled(IconButton)`
+  position: absolute;
+  background: linear-gradient(rgba(0, 0, 178, 1), rgba(0, 0, 80, 1));
+  top: 48px;
+  z-index: 2;
+  & svg {
+    color: #fff;
+  }
+`;
+
+function WrapperPage() {
+  const [xhoprValue, set_xHOPRValue] = useState('');
+  const [wxhoprValue, set_wxHOPRValue] = useState('');
+
+  // TODO:
+  const handleSwapClick = () => {
+    // const tempValue = xhoprValue;
+    // set_wxHOPRValue(xhoprValue);
+    // set_xHOPRValue('');
+  };
+  return (
+    <Section
+      lightBlue
+      center
+    >
+      <StyledPaper>
+        <h2>Wrapper</h2>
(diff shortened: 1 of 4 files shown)


src/sections/wrapper.tsx
```diff
@@ -0,0 +1,282 @@
...
+function WrapperPage() {
+  const [xhoprValue, set_xhoprValue] = useState('');
+  const [wxhoprValue, set_wxhoprValue] = useState('');
+  const [swapDirection, set_swapDirection] = useState<'xHOPR_to_wxHOPR' | 'wxHOPR_to_xHOPR'>('xHOPR_to_wxHOPR');
+  const { address } = useAccount();
+
+  // Fetch balance data
+  const { data: xHOPR_balance } = useBalance({
+    address,
+    token: '0xD057604A14982FE8D88c5fC25Aac3267eA142a08',
```
diegoalzate, line 129: const value for repeated sc addresses

diegoalzate, line 129: const value for repeated sc addresses

src/abi/wrapperAbi.json
```diff
@@ -0,0 +1,120 @@
...
+    "outputs": [
+      {
+        "name": "balance",
+        "type": "uint256"
+      }
+    ],
+    "payable": false,
+    "stateMutability": "view",
+    "type": "function"
+  },
```
diegoalzate, line 20: because we use `useBalance` hook we _can_ remove this. But can also stay for future possible purposes.

src/sections/wrapper.tsx
```diff
@@ -0,0 +1,282 @@
...
+function TransactionLink({
+  isSuccess,
+  hash,
+}: TransactionLinkProps) {
+  if (!isSuccess) return null;
+
+  return (
+    <span>
+      Check your{' '}
+      <GnosisLink href={`https://gnosisscan.io/tx/${hash}`}>
```
diegoalzate, line 112: open link on new page

src/sections/wrapper.tsx
```diff
@@ -0,0 +1,282 @@
...
+          </Button>
+          {(is_xHOPR_to_wxHOPR_loading || is_wxHOPR_to_xHOPR_loading) && <span>Check your Wallet...</span>}
+          <TransactionLink
+            isSuccess={is_xHOPR_to_wxHOPR_success}
+            hash={xHOPR_to_wxHOPR_data?.hash}
+          />
+          <TransactionLink
+            isSuccess={is_wxHOPR_to_xHOPR_success}
+            hash={wxHOPR_to_xHOPR_data?.hash}
+          />
```
diegoalzate, line 275: add more detail to be able to know what tx wrap it refers to

src/sections/wrapper.tsx
camilovegag: Added consts for smart contract addresses

src/abi/wrapperAbi.json
camilovegag: Leaving it, just in case

src/sections/wrapper.tsx
camilovegag: Done

src/sections/wrapper.tsx
camilovegag: Done

diegoalzate (approved):

mjadach-iv (changes requested):
Works.
Just please add a check for balances of xHOPR and wxHOPR on a successful swap. If there is someting in the wallet, update the fields on the wrapper.

mjadach-iv (changes requested):
After using it, I have following UX/UI thoughts and requests: 
 - on entering the page, the balances should be populated automatically with the balance of the wallet
 - the arrows do not show what direction will be the swap done, it should be a single arrow which onClick will turn around
 - it should be possible to swap the direction on clicking the oposite texfield, eg. if the xHOPR is highlighted and we click on wxHOPR, then the arror should turn around and go from wxHOPR to xHOPR and if you click again on the newly highlighted wxHOPR  textfield - nothing should happen

mjadach-iv (approved):

PR #87, opened 2023-07-03
Subpage: Ping

camilovegag:
### Tasks:
- [x] Create ping subpage

### DIFF
--- a/src/sections/ping.tsx
+++ b/src/sections/ping.tsx
@@ -9,7 +9,7 @@ import TextField from '@mui/material/TextField';
 function PingPage() {
   const dispatch = useAppDispatch();
   const [peerId, set_peerId] = useState('');
-  const [pinging, setPinging] = useState(false);
+  const [pinging, set_pinging] = useState(false);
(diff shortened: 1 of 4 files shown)


src/sections/ping.tsx
```diff
@@ -0,0 +1,113 @@
...
+      return !!aliases[alias];
+    } else return false;
+  };
+
+  const validatePeerId = (peerId: string) => {
+    if (isAlias(peerId)) {
+      if (aliases) {
+        return aliases[peerId];
+      } else return peerId;
+    } else return peerId;
```
diegoalzate, line 47:
```suggestion
    if (aliases && isAlias(peerId)) {
        return aliases[peerId];
    } 
    
    return peerId;
```

diegoalzate: I would do something like this, let me know if there are any issues with this logic

camilovegag: Sure! Will update on both sides we use this, it might be good to make this a util function in the future

diegoalzate (approved):

PR #88, opened 2023-07-03
Reconnect ws on page refresh

### LINKED ISSUE #65: Re connect ws after page refresh
## Expected Behavior

Web sockets should re-connect after page is refreshed

## Current Behavior

When you refresh the page `initializeMessagesWebsocket` and  `initializeLogsWebsocket` are not being called, we should have the `useNode` behavior if the apiToken and apiEnpoint are valid.

## (Optional) Possible Solution

Add the web socket initializers at the useEffect on `router.tsx`, maybe a code refactor would be needed? As we are working with React.StrictMode, this effect is being called twice but its not being unmounted, so if these actions are added right away we would get four web socket connections (duplicated ones)

## Steps to Reproduce

1. Login to node
2. Go to logs page
3. Refresh your browser
4. You still will be logged to the node, but you wont be able to see ant logs

diegoalzate:
fixes #65 

## overview
after refreshing the page ws connections are re-established.

### redux changes
now the ws status is not determined by a boolean but by a string type union, giving us the flexibility to add more statuses if needed but right now we can determine if a ws is connecting | connected | error

### DIFF
--- a/src/router.tsx
+++ b/src/router.tsx
@@ -203,6 +203,13 @@ const LayoutEnhanced = () => {
         apiEndpoint,
       }),
     );
+    dispatch(nodeActions.initializeMessagesWebsocket());
+    dispatch(nodeActions.initializeLogsWebsocket());
+
+    return () => {
+      dispatch(nodeActions.closeLogsWebsocket());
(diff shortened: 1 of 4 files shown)


camilovegag (approved):
Nice!

PR #92, opened 2023-07-04
import multiple channels function

### LINKED ISSUE #61: Admin subpage Channels: import multiple channels
blocked by hoprnet/hopr-sdk#38.

### Overview
As we can export channels as a `csv` file, we should also support importing multiple channels from a single click of a button.

The idea would be to parse the contents of the `csv` file into an array which would then be passed to re respective API call to open all the channels in the `csv` file.

However, specific handling of the `csv` import will be necessary, as the currently designed API implemented in the near future expects an array of only peerIds and the same amount to fund all channels.

### Definition of Done:

- Have a button that opens a popup window (Mui dialog is useful for this) that will ask to upload a csv file or manually input multiple peerIds and an amount to fund all the newly created channels.

ItsLuized:
Blocked by: hoprnet/hopr-sdk#71
Fixes #61 

### Overview

This pull request provides the functionality to upload a csv file that was previously exported from the channels subpage in order to open channels with all the peerIds present in that file.


diegoalzate (approved):
not sure about what happened to the yarn lock in this pr but the functionality works well

PR #95, opened 2023-07-04
migrate button from jsx to tsx

diegoalzate:
### overview
now button in future lib components is typed with tsx


camilovegag (approved):
lgtm

PR #97, opened 2023-07-05
Add safe pending transaction ui

### LINKED ISSUE #96: Pending safe signatures
Definition of done:
 - new subpage 'requires signature'
 - a list of pending transitions to sign in a table on that subpage
 - an action button in each row of the table to sign pending transitions
 - a list of pending transitions to sign in a notification bar 
 - when clicked on the pending transition in a notification bar, you get taken to the  'requires signature' subpage
 - a check on safe login if new pending transitions are to sign (watcher part?)
 - a check every 10s if new pending transitions are to sign (watcher part)

diegoalzate:
closes #96 

## overview
adds a section where users can sign/reject/confirm pending transactions.

todo:
- [x] pending txs ui
- [x] update watcher to add notifications when a new tx is added 

### How many transactions can a safe handle at a time?
Something that was new to me is that safe **does not** allow handling multiple transactions at the same time [source](https://help.safe.global/en/articles/40839-why-are-transactions-with-the-same-nonce-conflicting-with-each-other) unless they are created at the same time with [MultiSend](https://github.com/safe-global/safe-core-sdk/tree/main/packages/protocol-kit). 

### So what happens if I create a more than 1 pending proposal?
This could happen if you are very eager to push your proposal, however only one of those transactions will actually ever go through. Even if all the pending transactions have all approvals, only one can actually be executed. 

This is because after a transaction goes through the safe nonce gets bumped up by 1 and then all signatures that were previously created with the old nonce are basically useless. These useless transactions are actually just called rejected transactions [source](https://ethereum.stackexchange.com/a/119597).

### notable mentions in this pr
 - updated useSigner to useEthersSigner,  did this using the guide from [wagmi](https://wagmi.sh/react/ethers-adapters#wallet-client--signer) 
 - refactor useWatcher, separate the intervals from the business logic

### What state is my transaction in?
image

### things that can improve
The user after connecting to a safe will always receive a notification if there has a pending transaction. This logic can change in a future pr to only show notifications if the transaction was created after the user connected to the safe. Thought it was more useful to see a notification even if it was going to be repetitive. 

Also we could limit sending notifications to only users that have not signed, right now the notification will go to all owners from the safe regardless if they have signed or not.

### ui state
CleanShot 2023-07-06 at 19 24 08


mjadach-iv:
When switching safes a lot, the notifications do not get updated.

src/hooks/useSigner.ts
```diff
@@ -1,49 +1,30 @@
...
+    account,
+    chain,
+    transport,
+  } = walletClient;
 
-/**
- * React hook that gets signer from window.ethereum
- * @returns ethers signer
- */
-export const useSigner = () => {
```
mjadach-iv, line 12: That function seems to be still used in [hopr-admin](https://github.com/hoprnet/hopr-admin/tree/add-safe-signatures-required-section)/[src](https://github.com/hoprnet/hopr-admin/tree/add-safe-signatures-required-section/src)/[steps](https://github.com/hoprnet/hopr-admin/tree/add-safe-signatures-required-section/src/steps)/[install-node](https://github.com/hoprnet/hopr-admin/tree/add-safe-signatures-required-section/src/steps/install-node)
/nodeAddress.tsx

mjadach-iv (changes requested):

PR #100, opened 2023-07-06
Steps: create safe

camilovegag:
#### tl;dr
The issue we encountered was not related to the `owners` and `threshold` of the Safe contracts but rather the reuse of the salt nonce. By aligning our deployment process with the Safe contracts' timestamp-based salt nonce, we have resolved the problem and can proceed without further issues.

----

Today, we discovered an issue with the creation of Safe contracts from our UI. Upon inspecting the transaction data using an Ethereum call data decoder, we noticed that the deployed contracts were nearly identical to those created through the Gnosis Safe website. The only significant difference was the value of the `saltNonce`. While our UI was using a large string numerical value like `80072123504240615488077723526999251632250871316157101247713883127680525209488`, the Safe contracts had much smaller values like `1689178932310`.

After conducting thorough investigation with @diegoalzate, we identified that the problem originated from the `createProxyWithNonce` method used in the contract deployment process. This method employs the `CREATE2` opcode, which creates a contract with a deterministic address. Consequently, when attempting to create a Safe contract with identical owners and threshold through our UI, the operation failed.

The root cause of the issue was the reuse of the same salt and sender combination, which resulted in conflicting contract addresses. However, we realized that the Safe contracts used a timestamp as the `saltNonce`. To resolve the issue, we made the necessary modification to the `deploySafe` function, allowing it to generate a unique `saltNonce` based on the current timestamp. With this update in place, the contract creation process now functions correctly.

### Description:
To create, deploy, and fund a safe, follow these steps. 
You can add owners and select the threshold for your safe. Once you click "Continue," you'll have to wait a few seconds for the MetaMask popup to confirm the transaction. The `step` will change from 0 to 1, and the next screen will be rendered.

If you click the back button, the previous screen will be rendered, but the safe progress will be lost.

Here, you will add the amount of `xDAI` and `wxHOPR` you want to send to your safe and stake. I'm subtracting `0.003` from the max `xDAI` balance to account for gas fees.

Once you click "Deploy," you'll see a MetaMask popup for the `xDAI` transaction. If this transaction is successful, you'll then get a popup for the `wxHOPR`. Multicall was not an option here as it only supports [read methods](https://wagmi.sh/core/actions/multicall).

Finally, if everything goes well, the `step` will change from 1 to 2, and you'll see the safe deployed screen.
### Evidence:
Screenshot 2023-07-07 at 6 25 10 PM

image
image


src/sections/safe.tsx
```diff
@@ -1,196 +1,75 @@
...
-                }),
-              );
-              dispatch(
-                safeActionsAsync.getAllSafeTransactionsThunk({
-                  signer,
-                  safeAddress,
-                }),
-              );
-              dispatch(
-                safeActionsAsync.getSafeDelegatesThunk({
```
camilovegag, line 68: My changes removed this, where should this go?

mjadach-iv: There where it was before would be best.
Your new subpage call Onboarding for now

mjadach-iv (changes requested):
- when you create a safe, you need some kind of a loader to know that MM is doing something
- if you create and fund a safe, when you go back and do the process again, you do not get MM popup any more:
image

src/sections/safe.tsx
camilovegag: Created a new safe onboarding folder with the component, safe is now equal to develop branch

mjadach-iv (approved):

PR #122, opened 2023-07-13
variable to show only the parts of app we need

mjadach-iv:
const environment that can be populated with following strings: 'dev' | 'node' | 'web3'

'dev' will show all parts of the app
'node' will show only Node Admin parts of the app
'web3'will show only Staking Hub and Web3 parts of the app

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -1,0 +1,1 @@
+export const environment:('dev' | 'node' | 'web3') = 'dev';
--- a/src/router.tsx
+++ b/src/router.tsx
@@ -1,5 +1,6 @@
 import { useEffect } from 'react';
 import { createBrowserRouter, RouteObject, useSearchParams } from 'react-router-dom';
+import { environment } from '../config'
 
 // Store
 import { useAppDispatch, useAppSelector } from './store';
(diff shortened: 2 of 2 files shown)


config.ts
```diff
@@ -0,0 +1 @@
+export const environment:('dev' | 'node' | 'web3') = 'dev';
```
diegoalzate, line 1: Should this me able to be overwritten by environment variables ?

diegoalzate (changes requested):

config.ts
mjadach-iv: It is done on purpose like that, because when doing it in this way, during the build process the web-app is build without the components which it will not need. If we make the `const environment` into a variable, then we will have to ship the application with all the dependencies for both the Node Admin and Staking Hub

esterlus: Curious if this actually works.
Does the build process really remove all the other code if it is `const`?
Did you verify the output to only contain what you actually thinks it does?

Asking stupid questions because I remember this being a huge pain and never quite working as expected.

mjadach-iv: Yeah, its pretty cool as it really works with `vite` which we use currently.
Even if you will change the const to var, it will not change how it works.
Awesome thing is that is you chnage the code to 
```
export var environment:('dev' | 'node' | 'web3') = 'node';
setTimeout(()=>{environment = 'web3'}, 10000);
```
It will build the app for both 'node' and 'web3'.

diegoalzate: @esterlus just saw with @mjadach-iv and the bundle size of the project does vary. Quite interesting

diegoalzate (approved):

config.ts
mjadach-iv: Also, if you build it for 'web3' you will find `MetaMask-Emblem` in the main.js and for the 'node' you will not.

Issue #119, opened 2023-07-12
Transactions: Pending (styling)


mjadach-iv:
Definition of Done:
 - make them look like Historic Transactions

PR #126, opened 2023-07-13
Safe transactions increment nonce

### LINKED ISSUE #121: Safe transactions created in Staking Hub: increment nonce
Definition of Done:
- when creating any transation proposal in the Staking Hub, the nonce of the proposal should be bigger by one from the last transation proposal

diegoalzate:
closes #121 

### overview
Proposing a new tx will always now have a valid nonce

### example flow
1. safe nonce: 1
2. first proposed tx nonce: 1
3. second proposed tx nonce: 2
4. third proposed tx nonce: 3

### worth mentioning
This allows us to propose different transactions but the order is important. Since the safe nonce is 1 we should only interact with that one. 

The way rejections seem to work makes it impossible to reject a future nonce transaction.

##### weird flow interaction when interacting with future nonce tx
CleanShot 2023-07-13 at 14 57 48

we should disable future nonce interaction in #119

### DIFF
--- a/src/store/slices/safe/actionsAsync.ts
+++ b/src/store/slices/safe/actionsAsync.ts
@@ -206,10 +206,14 @@ const createSafeTransactionThunk = createAsyncThunk(
     try {
       const safeSDK = await createSafeSDK(payload.signer, payload.safeAddress);
       const safeApi = await createSafeApiService(payload.signer);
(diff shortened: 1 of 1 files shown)


mjadach-iv (approved):
Good job!

PR #139, opened 2023-07-18
Add low balance error on connect to node

### LINKED ISSUE #104: Connetct to node: low balance
Definition of done:
If unable to connect to the node, check the balance of the node:
- if the xDai balance will be below `0.001` then the `state.status.error` in authSlice should say `= 'Unable to connect.\n\nYour xDai balance seems to low to operate the node. Please top up your node.\nAddress: 0xXXXXXXXXXXXXXX \n\n' + JSON.stringify(meta.error);`
- if the xDai balance will be over `0.001` then leave the same logic as before

diegoalzate:
closes #104 

### overview
Adds a error message to auth state if node has less than 0.001 xDai

#### why in redux state?
There is already similar logic in loginThunk.rejected, so the main reason is to reuse already existing code.

### DIFF
--- a/src/store/slices/auth/actionsAsync.ts
+++ b/src/store/slices/auth/actionsAsync.ts
@@ -40,8 +40,7 @@ export const loginThunk = createAsyncThunk(
       const minimumNodeBalance = parseEther('0.001');
 
       if (nodeBalances?.native && BigInt(nodeBalances.native) < minimumNodeBalance) {
-        return rejectWithValue(`Unable to connect.
-        \n Your xDai balance seems to low to operate the node. 
+        return rejectWithValue(`Your xDai balance seems to low to operate the node. 
         \n Please top up your node.
         \n Address: ${addresses?.native}`);
       }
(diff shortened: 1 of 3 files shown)


src/store/slices/auth/actionsAsync.ts
```diff
@@ -1,16 +1,53 @@
...
+          apiToken,
+        }),
+      ).unwrap();
+
+      const addresses = await dispatch(
+        nodeActionsAsync.getAddressesThunk({
+          apiToken,
+          apiEndpoint,
+        }),
+      ).unwrap();
```
mjadach-iv, line 38: nice add

src/store/slices/auth/actionsAsync.ts
```diff
@@ -1,16 +1,53 @@
...
+
+      const minimumNodeBalance = parseEther('0.001');
+
+      if (nodeBalances?.native && BigInt(nodeBalances.native) < minimumNodeBalance) {
+        return rejectWithValue(`Unable to connect.
```
mjadach-iv, line 43: You already have `Unable to connect.` in the `loginThunk.rejected`.
Here it can be removed.

mjadach-iv (changes requested):
a minor change

mjadach-iv (approved):

PR #146, opened 2023-07-24
unix line breaks

mjadach-iv:


ItsLuized (approved):

mjadach-iv:
`yarn format` does only formatting, and `yarn format-lint` does formatting with linting, but doesn't shouw your errors
The `yarn lint` which I added shows the errors to you, and it was added to easily see what we will have to fix up
For now it is not needed, as none of the linting errors are critical and they do not affect built app

PR #147, opened 2023-07-24
Feature/safe threshold

### LINKED ISSUE #125: Safe: settings change safe threshold and owenrs
DEFINITION OF DONE:
- ui to change safe threshold. Should be some sort of dropdown similar to when we create a safe.
- redux actions to change safe threshold. Use createChangeThresholdTx from safe protocol kit

ItsLuized:
Fixes #125 

## Overview

This pull request adds the Safe Settings subpage which has the following functionality:

- Add and Remove Owners
- Change Safe Threshold, options displayed as a dropdown.


mjadach-iv (changes requested):
The remove owner seems to not do anything. The PR needs a 2nd look

PR #148, opened 2023-07-24
new connect to node modal

mjadach-iv:
- css fixes
- select and textfield update
- moved connect to node to a modal
- nicer styling of the connecting and connecting error


mjadach-iv:
conflicts for tomorrow

diegoalzate (changes requested):
Found a couple quirks in ux:
- when i connect to a node the url updates and then resets back to default url. So it goes something like this:

1. start with http://127.0.0.1:5173 
2. while connecting it will add the params apiToken and apiEndpoint
3. After connecting I lose the params and i get redirected to http://127.0.0.1:5173/node/info

- I also lose connection to the node i am connected to if i press `change node` and close the modal. So anytime i press change node I disconnect regardless from current node

diegoalzate (approved):
A small thing i saw which i'm not sure how it should be is even when the node fails to connect it sets the url search params for apiEndpoint and apiToken.

PR #152, opened 2023-07-24
set allowance for hopr token from safe

### LINKED ISSUE #107: Setting allowance for a channel
Docs: https://docs.openzeppelin.com/contracts/3.x/api/token/erc777#ERC777-allowance-address-address-

```
allowance(address holder, address spender) → uint256
approve(address spender, uint256 value) → bool
```

allowance (get)
 - holder: 
 - spender: 

approve (set)
 - spender: Hopr Channels
 - value: allowance value

diegoalzate:
closes #107 

### overview
This pr shows how we can interact with smart contracts from a safe account. It specifically adds the functionality to set allowance of hopr token to hopr channels from the safe account.

### why is `createApproveTransactionData` needed?
Since the allowance actually comes from the holder of tokens which in this case would be the Safe. This means we have to use safe interactions like `proposeTransaction` and `executeTransaction`. To use these interactions we have to pass a specific type `SafeTransactionDataPartial`, which amongst other things contain `{ to: string; value: string; data: string; operation?: OperationType }`. 

So we create the encoded data for the transaction and then pass the data for safe to execute.

[inspiration](https://ethereum.stackexchange.com/questions/116761/how-to-send-erc20-erc721-tokens-using-safe-core-sdk/116793#116793)

### ui
CleanShot 2023-07-24 at 16 07 10

this button can be seen in the /hub/safe subpage so any reviewers can test.

### DIFF
--- a/src/sections/safe.tsx
+++ b/src/sections/safe.tsx
@@ -10,6 +10,16 @@ import { useEthersSigner } from '../hooks';
 import { utils } from 'ethers';
 import { observePendingSafeTransactions } from '../hooks/useWatcher/safeTransactions';
 import { appActions } from '../store/slices/app';
+import { Address, encodeFunctionData, formatEther } from 'viem';
+import { erc20ABI, useContractRead } from 'wagmi';
+import { SafeTransactionDataPartial } from '@safe-global/safe-core-sdk-types';
+
+const HOPR_CHANNELS_SMART_CONTRACT_ADDRESS = '0xfabee463f31e39ec8952bbfb4490c41103bf573e'
+// Maximum possible value for uint256
+const MAX_UINT256 = BigInt(2 ** 256) - BigInt(1);
+// mHopr
+const HOPR_TOKEN_SMART_CONTRACT_ADDRESS = '0x66225dE86Cac02b32f34992eb3410F59DE416698'
+
 function SafeSection() {
   const dispatch = useAppDispatch();
   const safe = useAppSelector((store) => store.safe);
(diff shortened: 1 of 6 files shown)


src/sections/safe.tsx
```diff
@@ -2,14 +2,24 @@ import { useEffect, useState } from 'react';
...
 import Section from '../future-hopr-lib-components/Section';
 import { useEthersSigner } from '../hooks';
 import { utils } from 'ethers';
 import { observePendingSafeTransactions } from '../hooks/useWatcher/safeTransactions';
 import { appActions } from '../store/slices/app';
+import { Address, encodeFunctionData, formatEther } from 'viem';
+import { erc20ABI, useContractRead } from 'wagmi';
+import { SafeTransactionDataPartial } from '@safe-global/safe-core-sdk-types';
+
+const HOPR_CHANNELS_SMART_CONTRACT_ADDRESS = '0xfabee463f31e39ec8952bbfb4490c41103bf573e';
```
mjadach-iv, line 17: Please move all the addresses to the config.ts file in the root of the repo

src/sections/safe.tsx
```diff
@@ -201,6 +245,14 @@ function SafeSection() {
           ) : null}
         </div>
       ))}
+      <h2>approve hopr token to hopr channels</h2>
+      <span>allowance: {formatEther(BigInt(allowanceData?.toString() ?? '0'))}</span>
+      <button
+        disabled={!safe.selectedSafeAddress || !signer}
+        onClick={proposeApproveTransactionToSafe}
```
mjadach-iv, line 252: The function that we are calling - please move it to Redux and make it accept data it needs as payload

mjadach-iv (changes requested):

mjadach-iv (approved):

PR #154, opened 2023-07-25
Safe account in app bar not updating

### LINKED ISSUE #151: Safe account in appbar not changing when switching metamask account
### What is expected?
When i change metamask account, we have to reset the safe address because we are not sure if the new account is actually an owner of this safe. 

### What is actually happening?
The app bar does not change even though we are not connected to a safe account.

### How to recreate ?
1. Connect to MM
2. Connect to Safe
3. Switch to another account in MM, the bug is more apparent in pending transactions page

CleanShot 2023-07-24 at 15 49 32

diegoalzate:
closes #151 

### Overview
Fixes safe account in app bar not updating with redux app. Removed local state that tracked selectedSafeAddress and now it only uses redux state variable.

### how to test?
- connect to a safe
- change mm account
- the expected result is that the safe account in app bar should reset


ItsLuized (approved):
Tested and works wonders! 👍🏻

PR #157, opened 2023-07-25
Info status bar

### LINKED ISSUE #135: InfoStatus component
Definition of done:
 - components gets data from Redux (possibliy there is a need of getting more data into Redux)
 - if safe is not chosen, it looks like example 1
 - if safe is chosen, it looks like example 2

Figma: https://www.figma.com/file/qjTbRGulmsPBQ2E0aHY7HI/Staking-UI-v3?type=design&node-id=1280-7068&mode=design&t=iYg4FmDR1uMT6Lg2-0

image

ItsLuized:
Closes #135 

## Overview

This pull request adds the functionality of showing the balances of the wallet and safe for the `xDai`, `xHopr` and `wxHopr` tokens.

If no safe is selected, it will look like this:
image

When a safe IS selected, it will look like this:
image


mjadach-iv (changes requested):
Cool component.
Please place it in the InfoBar. You can change the `const drawer` into `const nodeDrawer` and create another one called `const we3Drawer` for the StakingHub.
Remember to change router.ts to make sure that rightDrawer will show up on connecting MM

ItsLuized:
When InfoBar is shown, even when node is not connected (Web3 info is shown), the space intended for the NodeDrawer is taken up, we should find a way to only take this space if the Node is connected.

mjadach-iv (approved):

Issue #158, opened 2023-07-26
Hopr Icon on navbar resets Redux states


ItsLuized:
### expected behavior
When clicking the Hopr icon on the navbar we should be redirected to the `home` page or the one that represents it.

### actual behavior
When we click the Hopr icon we get logged out of anything we are logged into (e.g: Web3 , Node, Safe as a consequence of disconnecting from Web3).

### how to recreate
1. Log in to Node, Web3, Safe or just any of the 2 that are independent of each other (Web3 and Node).
2. Click the Hopr Icon.

Demo vid:

![Hopr Icon Bug](https://github.com/hoprnet/hopr-admin/assets/32660456/e44d799d-8add-43bc-b24b-ee58747d1ef7)

PR #162, opened 2023-07-27
Info subpage update

mjadach-iv:
- Info subpage update
- not turning on balance watcher if not connected
- https://github.com/hoprnet/hopr-admin/issues/158
- Ticket subpage

### DIFF
--- a/src/components/InfoBar/index.tsx
+++ b/src/components/InfoBar/index.tsx
@@ -11,7 +11,7 @@ const drawerWidth = 160;
 interface Props {}
 
 const Contaier = styled(Box)`
-  display: none;  
+  display: none;
   @media screen and (min-width: 600px) {
     display: flex;
   }
--- a/src/components/InfoBar/index.tsx
+++ b/src/components/InfoBar/index.tsx
@@ -10,7 +10,7 @@ const drawerWidth = 160;
 
 interface Props {}
 
-const Contaier = styled(Box)`
+const Container = styled(Box)`
   display: none;
   @media screen and (min-width: 600px) {
     display: flex;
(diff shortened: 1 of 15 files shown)


src/components/InfoBar/index.tsx
```diff
@@ -4,12 +4,19 @@ import { useAppDispatch, useAppSelector } from '../../store';
...
 import styled from '@emotion/styled';
 import Toolbar from '@mui/material/Toolbar';
-import { ethers } from 'ethers';
+import { utils } from 'ethers';
 
 const drawerWidth = 160;
 
 interface Props {}
 
+const Contaier = styled(Box)`
```
camilovegag, line 13: small typo

src/future-hopr-lib-components/Navbar/navBar.tsx
```diff
@@ -138,13 +138,13 @@ const NavBar: React.FC<{
               <MenuIcon />
             </IconButton>
             <Logo className="logo-hopr">
-              <a href="/">
+              {/* <a href="/"> */}
```
camilovegag, line 141: state was lost due to `a` tag, maybe a Link from react router would solve this?

src/components/InfoBar/index.tsx
mjadach-iv: fixed

src/future-hopr-lib-components/Navbar/navBar.tsx
mjadach-iv: We can do it

camilovegag (approved):

Issue #161, opened 2023-07-27
Node: Update the node initial state and handling


mjadach-iv:
Definition of done:
- For every call we do in the Info subpage, add a 'reloading' boolean in redux so, the redux will know exactly when a call is being done. eg.:
InitialState:
```
  balances: {
    hopr: string | null;
    native: string | null;
    reloading: boolean;
  };
```

Reducer:
```
  builder.addCase(getBalancesThunk.pending, (state, action) => {
    if (action.payload) {
      state.balances.reloading = true;
    }
  });
  builder.addCase(getBalancesThunk.fulfilled, (state, action) => {
    if (action.payload) {
      state.balances = {
        native: action.payload.native,
        hopr: action.payload.hopr,
        reloading: false,
      };
    }
  });
  builder.addCase(getBalancesThunk.rejected, (state, action) => {
    if (action.payload) {
      state.balances.reloading = false;
    }
  });
  ```

- Update the Info page refresh to use the newely added states:
```
      <SubpageTitle
        title="Info"
        refreshFunction={fetchInfoData}
        reloading={balances.reloading || otherRefresh || otherRefresh}
      />
 ```

Issue #155, opened 2023-07-25
double notifications


diegoalzate:
### expected behaviour
No notification should come twice

### actual behaviour
Sometimes when connecting to a safe, or anything that triggers a change in the watcher. It could cause a condition where 2 fetch functions run at the same time and triggers the same notification twice.

### how to recreate ?
- press a safe in the app bar twice considerably fast

### possible solutions
- add in redux state a way to stop functions from running if they are still running. Effectively adding a debounce functionality.

PR #168, opened 2023-08-01
Change redux initial state to contemplate "isFetching" property

ItsLuized:
### Overview

In order to be able to manage when a redux state is being re-fetch we changed the structure of some initial states that are directly involved in a `GET` operation from only having the state information/data to an object consisting of data and a boolean property called `isFetching`

E.g:
- Previous address initial state:
  ```
  addresses: {
    hopr: string | null; native: string | null
  };
  ```
- Current initial state:
  ```
  addresses: {
    data: { hopr: string | null; native: string | null };
    isFetching: boolean;
  };
  ```

### Related Issues

#155 
#161


src/components/InfoBar/index.tsx
```diff
@@ -50,22 +50,22 @@ export default function InfoBar(props: Props) {
     <div>
       <AppBarFiller />
       <Title>Status</Title>
-      <Data>{info?.connectivityStatus}</Data>
+      <Data>{info.data?.connectivityStatus}</Data>
```
diegoalzate, line 53: This will re-render whenever is fetching changes even when the data does not change.

src/components/InfoBar/index.tsx
```diff
@@ -50,22 +50,22 @@ export default function InfoBar(props: Props) {
...
-      <Data>{balances?.native && utils.formatEther(balances.native)}</Data>
+      <Data>{balances.native && utils.formatEther(balances.native)}</Data>
 
       <Title>mHOPR balance</Title>
-      <Data>{balances?.hopr && utils.formatEther(balances.hopr)}</Data>
+      <Data>{balances.hopr && utils.formatEther(balances.hopr)}</Data>
 
       <Title>Peers seen in the network</Title>
-      <Data>{peers?.announced?.length || '-'}</Data>
+      <Data>{peers.data?.announced?.length || '-'}</Data>
```
diegoalzate, line 62: anytime `data` is used in the fe we can assume that it will re render whenever isFetching changes so lets try to keep .data out of the frontend

src/sections/aliases.tsx
```diff
@@ -67,11 +67,11 @@ function AliasesPage() {
       <button
         disabled={aliases !== null && Object.keys(aliases).length === 0}
         onClick={() => {
-          if (aliases) {
+          if (aliases.data) {
```
diegoalzate, line 70: there are a couple places where .data is being used, i think you can decide if which ones should be changed.

src/store/slices/node/actionsAsync.ts
```diff
@@ -559,113 +559,202 @@ const getPrometheusMetricsThunk = createAsyncThunk(
...
+    state.aliases.isFetching = false;
+  });
+  builder.addCase(getAliasesThunk.rejected, (state, action) => {
+    state.aliases.isFetching = false;
   });
   builder.addCase(getBalancesThunk.pending, (state, action) => {
     if (action.payload) {
-      state.balances.reloading = true;
+      state.balances.isFetching = true;
     }
```
diegoalzate, line 601: should this only change to true if there is a payload? what is the payload in this case?

src/store/slices/node/actionsAsync.ts
```diff
@@ -559,113 +559,202 @@ const getPrometheusMetricsThunk = createAsyncThunk(
...
   });
   builder.addCase(getBalancesThunk.fulfilled, (state, action) => {
     if (action.payload) {
       state.balances = {
         native: action.payload.native,
         hopr: action.payload.hopr,
-        reloading: false,
+        isFetching: false,
```
diegoalzate, line 608: This object does not follow the structure: {data: Object, isFetching: bool}

src/store/slices/node/actionsAsync.ts
```diff
@@ -559,113 +559,202 @@ const getPrometheusMetricsThunk = createAsyncThunk(
...
+  });
+  builder.addCase(getAliasThunk.pending, (state, action) => {
+    state.aliases.isFetching = true;
   });
   builder.addCase(getAliasThunk.fulfilled, (state, action) => {
-    if (action.payload && state.aliases) {
-      if (state.aliases) {
-        state.aliases[action.payload.alias] = action.payload.peerId;
+    if (action.payload && state.aliases.data) {
+      if (state.aliases.data) {
```
diegoalzate, line 734: I think there is redundancy here with `if (state.aliases.data)`

src/store/slices/node/actionsAsync.ts
```diff
@@ -559,113 +559,202 @@ const getPrometheusMetricsThunk = createAsyncThunk(
...
+  });
+  builder.addCase(getAliasThunk.rejected, (state, action) => {
+    state.aliases.isFetching = false;
   });
   builder.addCase(setAliasThunk.fulfilled, (state, action) => {
-    if (action.payload && state.aliases) {
-      if (state.aliases) {
-        state.aliases[action.payload.alias] = action.payload.peerId;
+    if (action.payload && state.aliases.data) {
+      if (state.aliases.data) {
```
diegoalzate, line 747: same as last comment

src/store/slices/node/actionsAsync.ts
```diff
@@ -807,13 +910,20 @@ export const createExtraReducers = (builder: ActionReducerMapBuilder<typeof init
   });
   builder.addCase(deleteTokenThunk.fulfilled, (state, action) => {
     if (action.payload?.deleted) {
-      state.tokens = state.tokens.filter((token) => token.id !== action.payload?.id);
+      state.tokens.data = state.tokens.data.filter((token) => token.id !== action.payload?.id);
     }
```
diegoalzate, line 914: seems like there is no update for isFetching here

src/store/slices/node/initialState.ts
```diff
@@ -26,22 +27,32 @@ export type Message = {
...
+    isFetching: boolean;
   };
-  aliases: GetAliasesResponseType | null;
+  aliases: { data: GetAliasesResponseType | null; isFetching: boolean };
   balances: {
     hopr: string | null;
     native: string | null;
-    reloading: boolean;
+    isFetching: boolean;
+  };
```
diegoalzate, line 51: i would move hopr and native into a data property

diegoalzate (changes requested):

src/store/slices/node/actionsAsync.ts
ItsLuized: Good catch, this should change without the payload.

src/store/slices/node/initialState.ts
```diff
@@ -26,22 +27,28 @@ export type Message = {
 type WebsocketConnectionStatus = 'connecting' | 'connected' | 'error' | null;
 
 type InitialState = {
-  info: GetInfoResponseType | null;
+  // info: GetInfoResponseType | null;
```
diegoalzate, line 30: remove comment

src/store/slices/node/actionsAsync.ts
```diff
@@ -559,122 +821,216 @@ const getPrometheusMetricsThunk = createAsyncThunk(
...
+const setPeersFetching = createAction<boolean>('node/setPeersFetching');
+const setPeerInfoFetching = createAction<boolean>('node/setPeerInfoFetching');
+const setEntryNodesFetching = createAction<boolean>('node/setEntryNodesFetching');
+const setSettingsFetching = createAction<boolean>('node/setSettingsFetching');
+const setStatisticsFetching = createAction<boolean>('node/setStatisticsFetching');
+const setTicketsFetching = createAction<boolean>('node/setTicketsFetching');
+const setTokensFetching = createAction<boolean>('node/setTokensFetching');
+const setVersionFetching = createAction<boolean>('node/setVersionFetching');
+const setTransactionsFetching = createAction<boolean>('node/setTransactionsFetching');
+
```
diegoalzate, line 840: this can maybe move somewhere else, seems to add noise and complexity here. We can tackle this in a different pr but i would say that this is not ideal

src/store/slices/node/initialState.ts
ItsLuized: removed!

src/store/slices/node/actionsAsync.ts
```diff
@@ -59,93 +81,207 @@ const {
...
+          error: e.error,
+        });
+      }
     }
-  }
-});
+  },
+  { condition: (_payload, { getState }) => {
+    const isFetching = getState().node.info.isFetching;
+    console.log('getInfo isFetching:', isFetching);
```
diegoalzate, line 105: i think we should remove this console log

src/store/slices/node/actionsAsync.ts
```diff
@@ -59,93 +81,207 @@ const {
...
+    if (isFetching) {
+      console.log('Cancelling getInfo request');
```
diegoalzate, line 107: same as above

src/store/slices/node/actionsAsync.ts
```diff
@@ -317,11 +615,31 @@ const removeAliasThunk = createAsyncThunk(
...
+  CloseChannelPayloadType,
+  { state: RootState }
+>(
   'node/closeChannel',
-  async (payload: CloseChannelPayloadType, { rejectWithValue }) => {
+  async (payload: CloseChannelPayloadType, {
+    rejectWithValue,
+    dispatch,
+  }) => {
+    dispatch(setChannelsFetching(true));
```
diegoalzate, line 642: This function can take time, if we set all channels to fetching then I won't be able to fund or get channels while this is happening. Makes all of these interactions exclusive. Not sure if we should separate this or if this makes sense.

cc @mjadach-iv

src/store/slices/node/actionsAsync.ts
diegoalzate: create an issue to make sure we don't forget this.

diegoalzate (changes requested):

src/store/slices/node/actionsAsync.ts
diegoalzate: This same pattern happens for tickets

src/store/slices/node/actionsAsync.ts
diegoalzate: Did not comment on every console log but you get the gist 😉

src/store/slices/node/actionsAsync.ts
ItsLuized: Comments removed from node and safe actionsAsync

src/store/slices/node/actionsAsync.ts
ItsLuized: Comments removed from node and safe actionsAsync

src/store/slices/node/actionsAsync.ts
ItsLuized: Issue made

ItsLuized:
The issue to move the actions away from the `actionsAsync` files is here: https://github.com/hoprnet/hopr-admin/issues/192

mjadach-iv (changes requested):
I do not like the .data key, it introduces a lot of unneded complexity.
I would rather if you extened the current objects with isFetching
@ItsLuized, @diegoalzate what do you think?

diegoalzate:
> I do not like the .data key, it introduces a lot of unneded complexity. I would rather if you extened the current objects with isFetching @ItsLuized, @diegoalzate what do you think?

I think extending the object makes it more complex on a typing level. I think this approach is also very similar to what other hook libraries do so it can make it easier to understand.

src/store/slices/node/actionsAsync.ts
ItsLuized: We added two new states: `closeChannel` and `redeemTickets` to aid this cases and handle them specifically

mjadach-iv (approved):

PR #176, opened 2023-08-07
Node Admin Style update, useBalance moved to Redux, all balances taken from redux, InfoBar style updated

mjadach-iv:
- Node Admin Style update,
-  useBalance moved to Redux, 
- all balances taken from redux, 
- InfoBar style updated
- changes Node Admin page titles to CAPS


src/components/InfoBar/index.tsx
```diff
@@ -122,44 +133,13 @@ export default function InfoBar(props: Props) {
...
-    token: wxhoprSmartContractAddress,
-    watch: true,
-  }).data?.formatted;
-  const safexHoprBalance = useBalance({
-    address: selectedSafeAddress as `0x${string}`,
-    token: xhoprSmartContractAddress,
-    watch: true,
-  }).data?.formatted;
+  const walletBalance = useAppSelector((store) => store.web3.balance);
+  const safeBalance = useAppSelector((store) => store.safe.balance);
```
ItsLuized, line 139: If one of the values in `safeBalance` and `walletBalance` is updated (e.g: xDai is updated from the safeBalance but not the other values) will it re-render the whole component? Should this be separated into different const values to avoid re-rendering if only one value is updated?

src/sections/wrapper.tsx
```diff
@@ -141,18 +146,7 @@ function WrapperPage() {
...
-    address,
-    token: xHOPR_TOKEN_SMART_CONTRACT_ADDRESS,
-    watch: true,
-  });
-  const { data: wxHOPR_balance } = useBalance({
-    address,
-    token: wxHOPR_TOKEN_SMART_CONTRACT_ADDRESS,
-    watch: true,
-  });
+  const walletBalance = useAppSelector((store) => store.web3.balance);
```
ItsLuized, line 149: Same comment as above, if one of the values comprehended inside the `walletBalance`, would it re-render the whole component?

src/steps/safeOnboarding/fundsToSafe.tsx
```diff
@@ -41,24 +40,18 @@ const FundsToSafe = ({
   set_step,
 }: FundsToSafeProps) => {
   const selectedSafeAddress = useAppSelector((store) => store.safe.selectedSafeAddress);
+  const walletBalance = useAppSelector((store) => store.web3.balance);
```
ItsLuized, line 43: Same comment as above, if one of the values comprehended inside the `walletBalance`, would it re-render the whole component?

ItsLuized (changes requested):
I just pointed to 3 parts where we use a variable for the whole state of `balance` for safe or web3, but if it is an improvement to separate into different variables, there's more instances

src/components/InfoBar/index.tsx
mjadach-iv: True, it will get rerendered again, but its small enough to leave it for now. We would need to close each part of the 6 balances into seperate components to avoid that

src/steps/safeOnboarding/fundsToSafe.tsx
mjadach-iv: Yes, good point but it will be a good thing to work on in a cople of months. 2

ItsLuized (approved):

PR #189, opened 2023-08-09
Styling changes: modals, notifications, overall colors

mjadach-iv:


src/future-hopr-lib-components/Layout/drawer.tsx
```diff
@@ -23,8 +23,8 @@ const StyledDrawer = styled(MuiDrawer)`
     padding-top: 59px;
     transition: width 0.4s ease-out;
     overflow-x: hidden;
-    background: #3c64a5;
-    color: white;
+    /* background: #3c64a5;
+    color: white; */
```
ItsLuized, line 27: we should remove the comments on the styled components.

ItsLuized (changes requested):
I found the issue with the scroll bar ( I think ) happening in the metamask menu when it's not connected.

A white line appears to the side of the right drawer
image

src/future-hopr-lib-components/Layout/drawer.tsx
mjadach-iv: Left it as I wanted to re-use it in next pr for the Staking Hub drawer. It's fixed now

mjadach-iv:
> I found the issue with the scroll bar ( I think ) happening in the metamask menu when it's not connected.
> 
> A white line appears to the side of the right drawer image

fixed

ItsLuized (approved):

PR #196, opened 2023-08-11
Move redeemTickets logic to redux

### LINKED ISSUE #164: Node: move handleRedeemAllTickets logic to Redux
Definition of done:
```

  const handleRedeemAllTickets = () => {
    set_redeeming(true);
    dispatch(
      actionsAsync.redeemTicketsThunk({
        apiEndpoint: loginData.apiEndpoint!,
        apiToken: loginData.apiToken!,
      }),
    )
      .unwrap()
      .then(() => {
        set_redeeming(false);
        set_redeemSuccess(true);
        handleRefresh();
      })
      .catch((e) => {
        set_redeeming(false);
        set_redeemSuccess(false);
        set_redeemErrors([
          ...redeemErrors,
          {
            error: e.error,
            status: e.status,
          },
        ]);
      });
  };
```
 - similar to above should be moved to Redux
 - Redux should know if redeming is in progress (boolean in Redux instead of  `const [redeeming, set_redeeming] = useState(false);`) 
 - redeemErrors should be also in Redux
 - connect ```
             <IconButton
              iconComponent={<ExitToAppIcon/>}
              tooltipText='Redeem all tickets'
              reloading={redeeming}
              onClick={handleRedeemAllTickets}
            />``` to redux redeeming state

ItsLuized:
Closes #164 

### Overview

This pull request aims to provide the functionality of handling when redeemTicketsThunk is loading and when it has errors in a global manner to reduce the amount of repeated amount of code.

The new redeemTickets state structure is:
```
redeemTickets: {
    isFetching: false,
    error: {
      error: undefined,
      status: undefined,
    },
  },
```

### DIFF
--- a/src/sections/node/tickets.tsx
+++ b/src/sections/node/tickets.tsx
@@ -50,6 +50,7 @@ function TicketsPage() {
         apiToken: loginData.apiToken!,
       }),
     );
+    handleRefresh();
   };
 
   return (
--- a/src/sections/node/tickets.tsx
+++ b/src/sections/node/tickets.tsx
@@ -20,9 +20,9 @@ function TicketsPage() {
   const ticketsFetching = useAppSelector((store) => store.node.tickets.isFetching);
   const statistics = useAppSelector((store) => store.node.statistics.data);
   const statisticsFetching = useAppSelector((store) => store.node.statistics.isFetching);
+  const redeemTicketsFetching = useAppSelector((store) => store.node.redeemTickets.isFetching);
+  const redeemTicketsErrors = useAppSelector((store) => store.node.redeemTickets.error);
   const loginData = useAppSelector((store) => store.auth.loginData);
-  const [redeemErrors, set_redeemErrors] = useState<{ status: string | undefined; error: string | undefined }[]>([]);
-  const [redeeming, set_redeeming] = useState(false);
 
   useEffect(() => {
     handleRefresh();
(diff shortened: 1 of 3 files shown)


src/store/slices/node/initialState.ts
```diff
@@ -98,7 +98,13 @@ type InitialState = {
   messagesWebsocketStatus: WebsocketConnectionStatus;
   logsWebsocketStatus: WebsocketConnectionStatus;
   closeChannel: { isFetching: boolean };
-  redeemTickets: { isFetching: boolean };
+  redeemTickets: {
+    isFetching: boolean;
+    error: {
+      status: string | undefined;
+      error: string | undefined;
+    };
```
diegoalzate, line 106: i think we can just have error: string

what do you think @mjadach-iv ?

diegoalzate (changes requested):

src/store/slices/node/initialState.ts
mjadach-iv: yeah, easier - better

mjadach-iv (changes requested):

src/sections/node/tickets.tsx
```diff
@@ -44,28 +44,12 @@ function TicketsPage() {
...
     dispatch(
       actionsAsync.redeemTicketsThunk({
         apiEndpoint: loginData.apiEndpoint!,
         apiToken: loginData.apiToken!,
       }),
-    )
-      .unwrap()
-      .then(() => {
-        set_redeeming(false);
-        handleRefresh();
```
diegoalzate, line 57: noticed that handleRefresh is no longer going to be called. Maybe not needed let me know.

diegoalzate (commented):
small question above

src/sections/node/tickets.tsx
ItsLuized: It is necessary, nice catch. I will move the handle refresh to happen after the reeemTickets

src/sections/node/tickets.tsx
```diff
@@ -44,28 +44,13 @@ function TicketsPage() {
...
-        set_redeemErrors([
-          ...redeemErrors,
-          {
-            error: e.error,
-            status: e.status,
-          },
-        ]);
-      });
+    );
+    handleRefresh();
```
diegoalzate, line 53: this will refresh before the action is finished. do .unwrap and then .then and inside that then you can refresh.

diegoalzate (changes requested):

PR #197, opened 2023-08-11
Fix error handling on connecting

### LINKED ISSUE #194: Node Admin: Error handling on connecting
Definition of Done:
 - Not enough balance shows up on the ERROR modal on connect to a node
 - websocket is not initialised if auth.status.connected is not true
 - This error doesn't change the erorr message from ERROR modal on connect to a node:
 ```
websocketMiddleware.ts:120 
 A non-serializable value was detected in an action, in the path: `payload`. Value: TypeError: Network request failed
    at xhr.onerror (browser-ponyfill.js:480:16) 
Take a look at the logic that dispatched this action:  {type: 'auth/login/rejected', payload: TypeError: Network request failed
    at xhr.onerror (http://127.0.0.1:5173/node_modules/.vite/deps…, meta: {…}, error: {…}}error: {message: 'Rejected'}meta: {arg: {…}, requestId: 'nB4cY7huzO4hTgHM9P0je', rejectedWithValue: true, requestStatus: 'rejected', aborted: false, …}payload: TypeError: Network request failed
    at xhr.onerror (http://127.0.0.1:5173/node_modules/.vite/deps/chunk-3XOKAOJW.js?v=f6827aa3:413:22)type: "auth/login/rejected"[[Prototype]]: Object 
(See https://redux.js.org/faq/actions#why-should-type-be-a-string-or-at-least-serializable-why-should-my-action-types-be-constants) 
(To allow non-serializable values see: https://redux-toolkit.js.org/usage/usage-guide#working-with-non-serializable-data)
```

### PR DESCRIPTION
**Pull request by diegoalzate**
closes #194 

### overview
changed the get loginThunk to catch more errors.

#### changes in redux
Sometimes i would see the error text get rendered twice mainly because the thunk would be in state `isFetching` and the redux condition would fail. So I added a `force` prop which would skip this condition.
```
{ condition: (_payload, { getState }) => {
    if (_payload.force) {
      return true;
    }

    const isFetching = getState().node.balances.isFetching;
    if (isFetching) {
      return false;
    }
  } },
```

### DIFF
--- a/src/hooks/useWatcher/balances.ts
+++ b/src/hooks/useWatcher/balances.ts
@@ -87,10 +87,10 @@ export const observeNodeBalances = ({
       if (!apiToken || !apiEndpoint) return;
 
       return await dispatch(
-        nodeActionsAsync.getBalancesThunk({ payload: {
+        nodeActionsAsync.getBalancesThunk({ 
           apiEndpoint,
           apiToken,
-        } }),
+        }),
       ).unwrap();
     },
     isDataDifferent: (newNodeBalances) => JSON.stringify(previousState) !== JSON.stringify(newNodeBalances),
--- a/src/hooks/useWatcher/balances.ts
+++ b/src/hooks/useWatcher/balances.ts
@@ -87,7 +87,7 @@ export const observeNodeBalances = ({
       if (!apiToken || !apiEndpoint) return;
 
       return await dispatch(
-        nodeActionsAsync.getBalancesThunk({ 
+        nodeActionsAsync.getBalancesThunk({
           apiEndpoint,
           apiToken,
         }),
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/hooks/useWatcher/balances.ts [hunk 1]
```diff
@@ -87,10 +87,10 @@ export const observeNodeBalances = ({
       if (!apiToken || !apiEndpoint) return;
 
       return await dispatch(
-        nodeActionsAsync.getBalancesThunk({
+        nodeActionsAsync.getBalancesThunk({ payload: {
```
mjadach-iv, line 90: Why is that change needed?

src/store/slices/node/actionsAsync.ts
```diff
@@ -114,11 +114,11 @@ const getAddressesThunk = createAsyncThunk<
       native: string;
     }
   | undefined,
-  BasePayloadType,
+  { payload: BasePayloadType; force?: boolean },
```
mjadach-iv, line 117: lets extend here BasePayloadType by force, it will be a lot easier to use the thunk later on

mjadach-iv (changes requested):

mjadach-iv (approved):

PR #202, opened 2023-08-14
Building process for production

mjadach-iv:
Tested on Windows and with Vercel: https://hub.hoprnet.org/
The extra scripts will change the config before build and bring it back after build.
We are not using env variables, as they are variables and not to make the final pakcage smaller.


scripts/post-build.js
```diff
@@ -0,0 +1,13 @@
...
+
+fs.unlink(configPath, function (err) {
+    if (err) { throw err }
+    else {
+        fs.rename(`${configPath}_`, configPath, function (err) {
+            if (err) throw err
+            console.log('Original config brought back.')
+        });
+    }
+});
```
diegoalzate, line 13: I'll create a seperate issue to remove `var` from here and see what we can do with eslint to make this file not throw linting errors.

diegoalzate (approved):
works

scripts/post-build.js
diegoalzate: created issue: https://github.com/hoprnet/hopr-admin/issues/208

PR #215, opened 2023-08-16
Infobar style update

mjadach-iv:
- styling of Infobar
- Trunkating moved mostly to css. For now only some whole nubmers will be trunkated by JS


ItsLuized (changes requested):
There is a gap between the info bar and the edge of the page to the right:
No wallet/safe connected
image

Wallet and safe connected:

image

mjadach-iv:
> There is a gap between the info bar and the edge of the page to the right: No wallet/safe connected image
> 
> Wallet and safe connected:
> 
> image

Looks as inteded

ItsLuized (approved):

PR #216, opened 2023-08-16
Tables with search, widthdraw modal in Info, 2 channel pages

mjadach-iv:
- Tables with search
- widthdraw modal in Info
- 2 channel pages

No need to make the tables with search better, as they will be redesgned to fit on mobile in the future


src/sections/node/channelsOutgoing.tsx
```diff
@@ -179,6 +180,66 @@ function ChannelsPage() {
       });
   };
 
+  const header = [
+    {
+      key: 'key',
+      name: 'key',
```
ItsLuized, line 186: Should we use the same name as in the incoming channels? (key is the same, but name is `#`)

src/sections/node/channelsOutgoing.tsx
```diff
@@ -179,6 +180,66 @@ function ChannelsPage() {
...
+    {
+      key: 'actions',
+      name: 'Actions',
+      search: false,
+      width: '168px',
+      maxWidth: '168px',
+    },
+  ];
+
+  const parsedTableData = Object.entries(channels?.incoming ?? []).map(([, channel], key) => {
```
ItsLuized, line 211: We are filtering on the incoming, but it should be outgoing.
```suggestion
  const parsedTableData = Object.entries(channels?.outgoing ?? []).map(([, channel], key) => {
```

ItsLuized (changes requested):
The texts in some tables are getting truncated and makes reading the info hard:

- The date is not shown completely and we have to hover over it to look at the complete date.
image

- Currency is not shown when large amount of decimals are present (we can infer it from the values below)
image

---
- The actions are not shown correctly, the last action's Icon is not distinguishable:
image

mjadach-iv:
> No need to make the tables with search better, as they will be redesgned to fit on mobile in the future

src/sections/node/channelsOutgoing.tsx
mjadach-iv: good catch

src/sections/node/channelsOutgoing.tsx
mjadach-iv: fixed

PR #237, opened 2023-08-24
Node admin: on login check NR

### LINKED ISSUE #230: Node Admin: On login, check for NR
Definition of Done:
On login, if the info will come back as `"isEligible": false?`, do not log-in to the dashboard, open the ERROR part of the pupup and info about not being registered similar to that one:

https://www.figma.com/file/qjTbRGulmsPBQ2E0aHY7HI/Staking-UI-v3?type=design&node-id=1550-17904&mode=design&t=VvaLu5GBXk38WYCR-0

diegoalzate:
closes #230 

### overview
checks if is eligible from info is true, if not it throws an error.

### DIFF
--- a/src/store/slices/auth/actionsAsync.ts
+++ b/src/store/slices/auth/actionsAsync.ts
@@ -23,6 +23,13 @@ export const loginThunk = createAsyncThunk<
       apiEndpoint: apiEndpoint,
       apiToken: apiToken,
     });
+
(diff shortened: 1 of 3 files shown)


mjadach-iv:
Approving, but the UI part of the task is not finished

mjadach-iv (approved):

PR #239, opened 2023-08-25
Onboarding 3

mjadach-iv:


src/components/Stepper/index.tsx
```diff
@@ -77,6 +77,7 @@ type StepProps = {
   name: string;
   position: number;
   state: StepState;
+  stepComplated: StepState;
```
diegoalzate, line 80: rename to lastStepCompleted

src/pages/staking-hub/onboarding.tsx
```diff
@@ -36,30 +33,56 @@ function Onboarding() {
...
+
+  function whatIsCurrentStep(page: number) {
+    switch (page) {
+    case 4:
+      return 1;
+
+    default:
+      return 0;
+    }
+  }
```
diegoalzate, line 55: this is still confusing. I think I would go for something like an object:

```
const STEPS =  {
 CREATE_SAFE: 1,
...
}
```

src/router.tsx
```diff
@@ -33,7 +33,7 @@ import UpdateNodePage from './steps/updateNode';
 import SetupNodePage from './steps/setupYourNode';
 import AddedToWhitelist from './steps/addedToWhitelist';
 import JoinWaitlistPage from './steps/joinWaitlist';
-import WhatYouWillNeedPage from './steps/whatYouWillNeed';
+import WhatYouWillNeedPage from './steps/step0_whatYouWillNeed';
```
diegoalzate, line 36: this comment is not specific to this exact file but the naming convention used in this file goes against what we have used in the rest of the repo so maybe we can have a folder step0 and inside that folder we can have that file

diegoalzate (commented):
some comments that we could branch into different prs if neccesary

src/router.tsx
mjadach-iv: folder is ok, @diegoalzate Diego please create a task to do on the topic, tbd when all steps are working

src/pages/staking-hub/onboarding.tsx
mjadach-iv: Good idea, will do sth like that in next PR

src/components/Stepper/index.tsx
diegoalzate: ok

src/router.tsx
diegoalzate: ok

diegoalzate (approved):

PR #251, opened 2023-08-28
Staking Hub: Enough Balance

ItsLuized:
### Overview

This pull request aim to add additional check to the add funds of the onboarding flow of the hopr-admin.

### What was done

- Add check to see if safe balance is enough and show green text to notify the user
- Enable 'continue' button is both token balances are enough
- Empty TextField after trx to fund is successful


src/pages/staking-hub/onboarding/step2/0fundSafe.tsx
```diff
@@ -91,6 +111,22 @@ const FundsToSafe = () => {
     write_wxHOPR_to_safe?.();
   };
 
+  const xdaiEnoughBalance = (): boolean => {
+    console.log(safeBalance.xDai.value);
+    if (Number(safeBalance.xDai.value) >= 0.1) {
```
mjadach-iv, line 116: value"200000000000000"
formatted:"0.0002"
gives `enough balance` although shouldn't

mjadach-iv, line 116: move the 0.1 and 1 values to the config.ts

mjadach-iv (changes requested):

src/pages/staking-hub/onboarding/step2/0fundSafe.tsx
ItsLuized: Good catch! This is fixed now

mjadach-iv (approved):

PR #266, opened 2023-08-28
Improve nft step

### LINKED ISSUE #255: Step NFT
1.	Community NFT: reset state in safe when changing selected Safe 
2.	OPTIONAL NFT TRANSFER: nft detedted in safe (figma)

diegoalzate:
closes #255

### overview
- shows success image on nft

state is already resetting when we connect to another safe so it is not included in this pr, this was handled in https://github.com/hoprnet/hopr-admin/pull/260


mjadach-iv (changes requested):
Rebase needed

vonlabs (approved):

PR #336, opened 2023-09-04
Switch network to Gnosis

mjadach-iv:
image


src/components/NetworkOverlay/index.tsx
```diff
@@ -1,114 +1,120 @@
 import React, { useState, useEffect, useRef } from 'react';
```
diegoalzate, line 1: not using any of these deps

mjadach-iv: fixed

mjadach-iv:
> When I switch to gnosis network the drawer items for `withdraw` and `dashboard` stay in disabled state.
> 
> Steps to recreate:
> 
> 1. go into dashboard on gnosis
> 2. switch to another network
> 3. switch back to gnosis using overlay button

fixed

PR #337, opened 2023-09-04
Landing page more mobile frieldly update

mjadach-iv:
Now: 
image

Before:
image


src/future-hopr-lib-components/Layout/index.tsx
```diff
@@ -70,6 +70,7 @@ const Layout: React.FC<{
   };
   drawerType?: 'blue' | 'white';
   drawerItems: ApplicationMapType;
+  drawerFunctionItems? : any;
```
diegoalzate, line 73:
```suggestion
  drawerFunctionItems? : ApplicationMapType;
```

diegoalzate: I think this is the type you use for this throughout the app right ?

diegoalzate (changes requested):
Small change and then looks good to merge

src/future-hopr-lib-components/Layout/index.tsx
mjadach-iv: Fixed

diegoalzate (approved):

PR #346, opened 2023-09-05
Update open channel modal placeholder

### LINKED ISSUE #343: Node Admin: Open channel modal
### overview
Open channel modal in peer subpage has a placeholder of peer id. This should peer address, as we now use the peer address to open channels

### definition of done
- [ ] modal shows correct placeholder and sends the correct information to the api

ItsLuized:
Closes #343 

## Overview

This PR corrects the modal placeholder and text to display 'peerAddress' instead of peerId.

### Note
Since the peer redux state does not have `peerAddress` yet, we left the field empty until this state becomes available


diegoalzate (approved):
Some formatting changes which could cause a problem in the future since we do not have a specific formatting file

PR #356, opened 2023-09-06
handle wallet errors if any

mjadach-iv:


src/components/ConnectWeb3/index.tsx
```diff
@@ -127,7 +146,15 @@ export default function ConnectWeb3({
...
-      set_localError(JSON.stringify(error)); 
+      let parsedError = '';
+      //@ts-ignore
+      if(error.shortMessage && error.shortMessage.length > 10) parsedError = error.shortMessage; 
+      //@ts-ignore
+      if(error.details && error.details.length > 10) parsedError = parsedError + '\n\n' + error.details;
+      if(parsedError === '') parsedError = JSON.stringify(error); 
+
+
+      set_localError(parsedError);
```
diegoalzate, line 157: I think this error handling would be better if typed. The error that I believe we are trying to catch is `UserRejectedRequestError`. So it could look something like this:

```
    if (error instanceof UserRejectedRequestError) {
        let parsedError = error.shortMessage; 
        if(error.details && error.details.length > 10) {
          parsedError = parsedError + '\n\n' + error.details;
        }
        set_localError(parsedError); 
      } else {
        set_localError( JSON.stringify(error))
      }
```

or even 
```
    if (error instanceof UserRejectedRequestError) {
    // we maybe can hardcode since we know what type of error this is
        set_localError('User rejected the connection'); 
      } else {
        set_localError( JSON.stringify(error))
      }
```

diegoalzate (changes requested):

diegoalzate (approved):
nice

PR #363, opened 2023-09-07
Node/update

mjadach-iv:


src/components/Modal/node/SendMessageModal.tsx
```diff
@@ -39,16 +39,17 @@ export const SendMessageModal = ({ peerId }: SendMessageModalProps) => {
...
+  const [directMessage, set_directMessage] = useState<boolean>(true);
   const [message, set_message] = useState<string>('');
   const [receiver, set_receiver] = useState<string>(peerId ? peerId : '');
   const [openModal, set_openModal] = useState<boolean>(false);
 
   const maxLength = 500;
   const remainingChars = maxLength - message.length;
 
-  const nonAutomaticPathTooltip = 'Disable `automatic path` to enable `Number of hops`';
+  const nonAutomaticPathTooltip = 'Disable `direct message` and `automatic path` to enable this field';
```
diegoalzate, line 52: when direct message is enabled, this tooltip does not show up

src/future-hopr-lib-components/Modal/styled.tsx
```diff
@@ -1,11 +1,19 @@
 import styled from '@emotion/styled';
 import { Dialog, DialogContent, IconButton } from '@mui/material';
 
-export const SDialog = styled(Dialog)`
+export const SDialog = styled(({
+  width,
+  ...rest
+}: any) => <Dialog {...rest} />)`
```
diegoalzate, line 7: We should change this any in either this pr or another, the code does work but we should try to minimize the use of anys

diegoalzate (approved):

src/components/Modal/node/SendMessageModal.tsx
mjadach-iv: Thx, updated

src/future-hopr-lib-components/Modal/styled.tsx
mjadach-iv: updated

PR #364, opened 2023-09-07
Add AliasByPeerAddress

ItsLuized:
Related #358

### Overview
In the last changes to the sdk we were able to get peerAddress and peerId when querying the peers that are announced on the network. With this new addition we can bring back the functionality of showing the alias of the node that has a channel open from or to us.

### DIFF
--- a/src/pages/node/channelsIncoming.tsx
+++ b/src/pages/node/channelsIncoming.tsx
@@ -13,7 +13,6 @@ import IconButton from '../../future-hopr-lib-components/Button/IconButton';
 import TablePro from '../../future-hopr-lib-components/Table/table-pro';
 
 // Modals
-import { PingModal } from '../../components/Modal/node/PingModal';
 import { OpenOrFundChannelModal } from '../../components/Modal/node/OpenOrFundChannelModal';
 
 // Mui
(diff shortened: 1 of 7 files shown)


src/pages/node/channelsIncoming.tsx
```diff
@@ -127,7 +144,7 @@ function ChannelsPage() {
     return {
       id: channel.id,
       key: key.toString(),
-      peerAddress: channel.peerAddress,
+      peerAddress: getAliasByPeerAddress(channel.peerAddress),
       status: channel.status,
       funds: `${utils.formatEther(channel.balance)} ${HOPR_TOKEN_USED}`,
       actions: (
```
diegoalzate, line 150: We would also want to activate the ping modal again, so i guess getAliasOrPeerIdFromPeerAddress is needed

diegoalzate (changes requested):

src/pages/node/channelsIncoming.tsx
ItsLuized: added getPeerIdFromPeerAddress to make the ping work

src/pages/node/channelsOutgoing.tsx
```diff
@@ -234,8 +242,7 @@ function ChannelsPage() {
       funds: `${utils.formatEther(channel.balance)} ${HOPR_TOKEN_USED}`,
       actions: (
         <>
-          {/* we need a way to get peer id to ping from channel */}
-          {/* <PingModal peerId={channel.peerId} /> */}
+          <PingModal peerId={getPeerIdFromPeerAddress(channel.peerAddress)} />
```
diegoalzate, line 245: Also do this on incoming channels

diegoalzate (changes requested):

src/pages/node/channelsOutgoing.tsx
ItsLuized: Nice catch! Thanks

PR #368, opened 2023-09-07
Feedback on chain component

### LINKED ISSUE #354: Staking Hub: Feedback on blockchain transaction components
### overview
There should be a consistent feed back component that can handle different states and work in a way it is re usable 

### ui example
CleanShot 2023-09-06 at 08 00 03@2x

The copy is not accurate, the copy we have right now is better.

diegoalzate:
closes #354 

### overview
created a `FeedbackTransaction` that receives 

```
{
  transactionHash?: Address;
  confirmations: number;
  feedbackTexts: {
    loading: string;
    idle?: string;
    success?: string;
    error?: string;
  };
}
```

Depending on the status of the transaction hash it will show the feedback text. Confirmation number allows customization on how long we would want to wait to make sure a tx went through.

### ui
CleanShot 2023-09-07 at 16 20 29


config.ts
```diff
@@ -27,8 +27,8 @@ export const STAKING_V2_SUBGRAPH = 'https://stake.hoprnet.org/api/hub/subgraph-d
 export const VITE_WALLET_CONNECT_PROJECT_ID = 'efdce6b5c6b10913211ff1b40bc4d54d';
 
 // Minimum to be funded
-export const MINIMUM_WXHOPR_TO_FUND = 30_000;
+export const MINIMUM_WXHOPR_TO_FUND = 0.03;
 export const MINIMUM_WXHOPR_TO_FUND_NFT = 10_000;
-export const MINIMUM_XDAI_TO_FUND = 2;
-export const MINIMUM_XDAI_TO_FUND_NODE = 1;
+export const MINIMUM_XDAI_TO_FUND = 0.02;
+export const MINIMUM_XDAI_TO_FUND_NODE = 0.01;
```
mjadach-iv, line 33: Need to revert that

mjadach-iv (changes requested):
Please use this component in all places that require wallet confirm

diegoalzate:
> Please use this component in all places that require wallet confirm

Added to fund node but did not add it to sign functions because we do not wait on those. Also reverted the minimum values.

mjadach-iv (changes requested):
- not present on 'staking/withdraw' path
- not present on 'staking/stake-wxHOPR' path
- not present on 'staking/fund-node' path

mjadach-iv (changes requested):
Still doesn't show on some components, eg:
image

diegoalzate:
> Still doesn't show on some components, eg: image

added a isWalletPendingState for this use case

mjadach-iv (changes requested):
Right now the `/staking/withdraw` is jumping when we press 'Execute' because the Feedback component is showing up. It also might be the case on other components.
It would be best to make the component have min-height of its normal height when displayed, so when the status would show, nothing would be moving around on the scren

PR #369, opened 2023-09-07
Add initial #staking hash if no hash is provided

### LINKED ISSUE #353: Add #staking to the url in the Dashboard: Staking so the FAQ will show up on 1st load

ItsLuized:
Fixes #353 

### Overview
When a client enters the dashboard, the initial screen that will appear will be the staking screen, although this works as intended, the FAQ component doesn't show because the `#staking` in the url is not present. This pull request adds said string to the url if the dashboard url doesn't have any `#` already.


ItsLuized:
Found a bug: if we mannually enter the  dashboard url and put #node, it will still show the staking screen

mjadach-iv (approved):

PR #396, opened 2023-09-12
Add force login to node admin

### LINKED ISSUE #391: Login Anyways - orange button
Definition of Done:
On the Error page of the login to the Node Admin, add an button with orange color with text 'LOGIN ANYWAYS' with a tooltip: 'Your did not start properly and might not be fully functioning. Some features might be offline and not working.'. On click of that button, a new modal should be opened with the same text, title 'WARNING' , Back button and 'LOGIN ANYWAYS' button, similar/based of this one:

image

After clicking 'LOGIN ANYWAYS'  on that 2nd modal, you let a user into the Node Admin. We do not check here if things do not work as intended.

diegoalzate:
closes #391 

### ui on error
CleanShot 2023-09-12 at 11 09 07

### ui on force login
image

why wip?
- missing a max-height property on modals that allows modal to grow with text

### DIFF
--- a/src/components/ConnectNode/modal.tsx
+++ b/src/components/ConnectNode/modal.tsx
@@ -378,7 +378,7 @@ function ConnectNodeModal(props: ConnectNodeModalProps) {
           </Button>
         </ConnectContainer>
 
-        {error && forceLogin && (
+        {error && (
           <Overlay className={'overlay-has-error'}>
             <CloseOverlayIconButton
               color="primary"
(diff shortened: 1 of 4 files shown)


mjadach-iv:
Did the styling, but I still can't force connect:
image

diegoalzate:
@mjadach-iv found a bug in current implementation. We have to be able to know when we wan to force and when we basically can not (Authentication error).

I was thinking about splitting these types of errors to FETCH_ERROR (<-- i think we can force node admin when fetch fails), API_ERROR (this is an error sdk has caught it includes authentication errors) and NOT_ELIGIBLE_ERROR ( this is one of those errors we create and we can force login)

Basically, force login when we know the user can authenticate and not checking if they are eligible.

mjadach-iv:
> @mjadach-iv found a bug in current implementation. We have to be able to know when we wan to force and when we basically can not (Authentication error).
> 
> I was thinking about splitting these types of errors to FETCH_ERROR (<-- i think we can force node admin when fetch fails), API_ERROR (this is an error sdk has caught it includes authentication errors) and NOT_ELIGIBLE_ERROR ( this is one of those errors we create and we can force login)
> 
> Basically, force login when we know the user can authenticate and not checking if they are eligible.

Ok, good idea - go for it

mjadach-iv (approved):

Issue #389, opened 2023-09-11
Withdraw NFT From Safe


mjadach-iv:
Definition of Done:
 - add new option to withdraw page http://localhost:5173/staking/withdraw?token=NFT
 - the option in select should have disabled the entries which are not present in the safe
 - when NFT is chosen, the currecny should be changed to 'id'
 - the id of the nft should be shown in the last texfield
 - add a GrayCard to  `http://localhost:5173/staking/dashboard#safe` with image 'NFT enabled' 'NFT not present' images (on the right) and 'tranfer' button on the left (disabled if nft not present)
 - https://gnosisscan.io/token/0x43d13d7b83607f14335cf2cb75e87da369d056c7  - HOPR Boost NFT
 - transfer function : safeTransferFrom

PR #397, opened 2023-09-12
Add withdraw NFT

diegoalzate:
closes https://github.com/hoprnet/hopr-admin/issues/389

### ui in safe
CleanShot 2023-09-12 at 16 35 21

### ui in withdraw
CleanShot 2023-09-12 at 16 35 44


mjadach-iv (changes requested):

diegoalzate:
@mjadach-iv found a bug in current implementation. We have to be able to know when we wan to force and when we basically can not (Authentication error).

I was thinking about splitting these types of errors to `FETCH_ERROR` (<-- i think we can force node admin when fetch fails), `API_ERROR` (this is an error sdk has caught it includes authentication errors) and `NOT_ELIGIBLE_ERROR` ( this is one of those errors we create and we can force login)

Basically, force login when we know the user can authenticate and not checking if they are elegible.

PR #405, opened 2023-09-14
Adding x86 & x64 support to the docker command

Jaguaras:


vonlabs (changes requested):
Will not get copied, also need to change the line 132

mjadach-iv (approved):

PR #410, opened 2023-09-14
peerAddress to nodeAddress and fix open multiple channels

mjadach-iv:
closes https://github.com/hoprnet/hopr-admin/issues/398

PR #418, opened 2023-09-14
Turn off wagmi when Node Admin

mjadach-iv:
Need to merge https://github.com/hoprnet/hopr-admin/pull/410 1st


mjadach-iv:
Fixed in https://github.com/hoprnet/hopr-admin/pull/611

PR #424, opened 2023-09-15
Sidebar state to localStorage

### LINKED ISSUE #404: Save in localstorage if user opened or closed the sidebar to later use that value

ItsLuized:
Fixes #404 

### Overview
This pull request saves to localStorage the last state of the Sidebar, either open or closed. When the page loads it loads the state from localStorage before setting the initial state.

### DIFF
--- a/src/future-hopr-lib-components/Layout/index.tsx
+++ b/src/future-hopr-lib-components/Layout/index.tsx
@@ -15,6 +15,7 @@ import Drawer from './drawer';
 // Types
 import { ApplicationMapType } from '../../router';
 import { useAppSelector } from '../../store';
+import { loadStateFromLocalStorage, saveStateToLocalStorage } from '../../utils/localStorage';
 
 const SLayout = styled.div`
   &.webapp {
@@ -70,7 +71,7 @@ const Layout: React.FC<{
   };
   drawerType?: 'blue' | 'white';
   drawerItems: ApplicationMapType;
-  drawerFunctionItems? : ApplicationMapType;
+  drawerFunctionItems?: ApplicationMapType;
   drawerRight?: React.ReactNode;
 }> = ({
   className = '',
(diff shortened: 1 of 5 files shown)


src/utils/localStorage.ts
```diff
@@ -0,0 +1,8 @@
+export function saveStateToLocalStorage(key: string, state: boolean) {
+    localStorage.setItem(key, JSON.stringify(state));
+}
+
+export function loadStateFromLocalStorage(key: string): string | null {
+    const storedState = localStorage.getItem(key);
+    return storedState ? JSON.parse(storedState) : null;
+}
```
diegoalzate, line 8: there is something similar to this already in utils/functions. I like what you did here but remove the function from the other place just for this to be the only place for it to be. Also `saveStateToLocalStorage` should receive a state that is unknown. That way it can work for other use cases.

diegoalzate (changes requested):

src/utils/localStorage.ts
ItsLuized: Fixed! I like the unknown for the state type 👍🏻

src/future-hopr-lib-components/Layout/index.tsx
```diff
@@ -85,56 +86,68 @@ const Layout: React.FC<{
   drawerType,
   drawerFunctionItems,
 }) => {
-  const isMobile = !useMediaQuery('(min-width: 500px)');
-  const isConnected = useAppSelector((store) => store.auth.status.connected);
+    const isMobile = !useMediaQuery('(min-width: 500px)');
+    const isConnected = useAppSelector((store) => store.auth.status.connected);
+    const initialDrawerState = loadStateFromLocalStorage('drawerState');
```
diegoalzate, line 91: maybe we should move this to a redux state, might make loading better

diegoalzate (approved):
works well

PR #426, opened 2023-09-15
Add option to disable notifications

### LINKED ISSUE #413: Node Admin: Notifications disable
Definition of Done:
- Add an option to turn off notifications for all types of notifications ( list of types with swiches next to them)
- It should in the Configuration page
- should be savedto localstorage and redux
- on load of the add, get data  from localstorage and save it to redux

diegoalzate:
closes #413 

### Overview
Adds a `active` prop to all watchers that allows to deactivate the fetch mechanism. So the interval runs but it does not do anything.

### ui
image

### DIFF
--- a/src/pages/node/configuration.tsx
+++ b/src/pages/node/configuration.tsx
@@ -11,12 +11,23 @@ import Button from '../../future-hopr-lib-components/Button';
 
 // Mui
 import { Paper, Switch, SelectChangeEvent } from '@mui/material';
+import styled from '@emotion/styled';
+import { appActions } from '../../store/slices/app';
+
+
+const NotificationsContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 1rem;
+`
 
 function SettingsPage() {
   const dispatch = useAppDispatch();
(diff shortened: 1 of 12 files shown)


src/pages/node/configuration.tsx
```diff
@@ -166,6 +187,83 @@ function SettingsPage() {
...
+                      onChange={() => {
+                        if (localNotificationSettings) {
+                          set_localNotificationSettings({
+                            ...localNotificationSettings, pendingSafeTransaction: !localNotificationSettings.pendingSafeTransaction,
+                          })
+                        }
+                      }}
+                      color="primary"
+                    />{' '}True
+                  </div>
```
mjadach-iv, line 263: Do we get those in the Node Admin?

mjadach-iv (changes requested):

src/pages/node/configuration.tsx
diegoalzate: probably not because they need to have a selectedSafe.

mjadach-iv (approved):

PR #435, opened 2023-09-19
Update sendMessage modal style

### LINKED ISSUE #431: Message Modal: improve Send mode
Definition of done:
 - send mode should be a select ocupying 50% of the modal in terms of width
 - if Number of hops is chosen, then in the right side the number of hops textfield should show:
image
- if path is chosen, then a texfield similar to the one in 'import .csv channels to open' should show up under the select:
image
- the path textfield should automatically enter a new line if a 'space' is pressed (instead of entering space in the field)

Unfortunetely the modal will jump, but unless there is abetter solution without opening up new popups, we need to stick with it for now

ItsLuized:
Fixes #431 

### Overview

This pull request introduces visual improvementes to the sendMessage Modal

image
image
image
image
---
#### The select:
image

### DIFF
--- a/src/components/Modal/node/SendMessageModal.tsx
+++ b/src/components/Modal/node/SendMessageModal.tsx
@@ -6,6 +6,7 @@ import { SDialog, SDialogContent, SIconButton, TopBar } from '../../../future-ho
 import Checkbox from '../../../future-hopr-lib-components/Toggles/Checkbox';
 import IconButton from '../../../future-hopr-lib-components/Button/IconButton';
 import Button from '../../../future-hopr-lib-components/Button';
+// import Select from '../../../future-hopr-lib-components/Select';
 
 // Mui
 import {
@@ -13,7 +14,10 @@ import {
   DialogActions,
   CircularProgress,
   TextField,
-  Tooltip
+  Tooltip,
+  SelectChangeEvent,
+  Select,
+  MenuItem
 } from '@mui/material'
 
 import { SendMessagePayloadType } from '@hoprnet/hopr-sdk';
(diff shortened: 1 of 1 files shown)


mjadach-iv (changes requested):
- For the Number of hops -> 0 hop mesasge, enable the sand button and the the message using direct path
- For Path option, update the helpertext to have the new peerId types starting with 12, and also make it have new lines after commas to be more clear

mjadach-iv (changes requested):
0 hop doesnt work:
{"tag":1,"body":"test2","peerId":"12D3KooWFSxmWxKLVRBnUb8G76Scd9ZTLYtBZjLmgPz429LJbArA","path":[],"hops":0}

direct works:
{"tag":1,"body":"test1","peerId":"12D3KooWFSxmWxKLVRBnUb8G76Scd9ZTLYtBZjLmgPz429LJbArA","path":[]}

src/components/Modal/node/SendMessageModal.tsx
```diff
@@ -147,14 +159,29 @@ export const SendMessageModal = (props: SendMessageModalProps) => {
...
+      const input = e.target as HTMLInputElement;
+      // These are always a number... never null
+      const start = input.selectionStart!;
+      const end = input.selectionEnd!;
+      set_path((prevPath) => prevPath.substring(0, start) + '\n' + prevPath.substring(end));
+      // Set a timeout to ensure that the focus and blur events have a chance to complete
+      setTimeout(() => {
+        input.setSelectionRange(start + 1, start + 1);
+      }, 0);
+    }
```
mjadach-iv, line 182:
```suggestion
    else if (e.key === ',' ) {
      e.preventDefault(); // Prevent space key from inserting a space
      const input = e.target as HTMLInputElement;
      // These are always a number... never null
      const start = input.selectionStart!;
      const end = input.selectionEnd!;
      set_path((prevPath) => prevPath.substring(0, start) + ',\n' + prevPath.substring(end));
      // Set a timeout to ensure that the focus and blur events have a chance to complete
      setTimeout(() => {
        input.setSelectionRange(start + 1, start + 1);
      }, 0);
    }
```

mjadach-iv: also add new line for `,`
This should work, please test

mjadach-iv (changes requested):

src/components/Modal/node/SendMessageModal.tsx
ItsLuized: Selection range had to be adjusted, but good for everything else! 👍🏻

mjadach-iv (approved):

PR #445, opened 2023-09-25
Change nginx docker version + add run command on README

ItsLuized:
### Overview

- While testing the docker image generated from the command, it was noticed that there were some io_setup() related errors. To fix them the `nginx` image was changed to the stable version.
- The docker run command was also added to the README

### DIFF
--- a/scripts/NodeAdmin.Dockerfile
+++ b/scripts/NodeAdmin.Dockerfile
@@ -27,7 +27,7 @@ COPY . .
 
 RUN yarn run build-node
 
-FROM nginx:1.25.2@sha256:48a84a0728cab8ac558f48796f901f6d31d287101bc8b317683678125e0d2d35 AS runtime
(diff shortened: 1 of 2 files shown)


scripts/NodeAdmin.Dockerfile
```diff
@@ -27,7 +27,7 @@ COPY . .
 
 RUN yarn run build-node
 
-FROM nginx:1.25.2@sha256:48a84a0728cab8ac558f48796f901f6d31d287101bc8b317683678125e0d2d35 AS runtime
+FROM nginx:stable AS runtime
```
mjadach-iv, line 30: lets use a specific stable version here that works, so it will not get damaged again

mjadach-iv (changes requested):

mjadach-iv (approved):

PR #447, opened 2023-09-28
Add better error message when get_peer_multiaddresses fails

### LINKED ISSUE #409: Better error when Node is starting
1. When node is starting, the info API call says that it can not get multiaddress. 
Make the ERROR message shown on the ERROR popup say in that case: 
`You Node seems to be starting, wait a couple of minutes before accessing it.` 
and after in a new line add the actual error message

2. On clicking 'login' get address right away as well as info, and always show the address on the ERROR popup below the error like: `Node address: 0xXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX`. The orange button should be under this line.
When we will be getting address right away, then there is no need to get them after login or if the node balance is low.

Now:
image

resp:
`{
    "status": "UNKNOWN_FAILURE",
    "error": "Cannot read properties of undefined (reading 'get_peer_multiaddresses')"
}`

ItsLuized:
Fixes #409 

# Overview

This pull request adds a more descriptive error message for when the node is still starting and it's not ready to be connected to.

- [x] When node is starting, the info API call says that it can not get multiaddress. Make the ERROR message shown on the ERROR popup say in that case:
        `You Node seems to be starting, wait a couple of minutes before accessing it and after in a new line add the actual error message`

- [x] On clicking 'login' get address right away as well as info, and always show the address on the ERROR popup below the error like: Node address: 0xXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX. The orange button should be under this line.
When we will be getting address right away, then there is no need to get them after login or if the node balance is low.


ItsLuized:
## Updated error message with node native address
image

mjadach-iv (approved):

PR #450, opened 2023-09-29
Add filter to remove consecutive or trailing commas

### LINKED ISSUE #449: Message modal
If there is a comma at the end of the path, remove it so the message doesnt fail
Best way will be to filter out from the parsed path array the nulls and empty strings

ItsLuized:
Fixes #449 

# Overview

Remove trailing or consecutive commas. Example problem messagePayload:
```
{
    “apiToken":"Example-api-token",
    "apiEndpoint":"http://apiEndpoint:3001/",
    "body":"ejemplo",
    "peerId":"ejemplo",
    "tag":1,
    "path":[
        "",
        "abc",
        "",
        "def",
        "cba”,
        ””
    ]
}
```

How it looks before sending to sdk (after filtering):
```
{
    "apiToken": "Example-api-token",
    "apiEndpoint": "http://apiEndpoint:3001/",
    "body": "ejemplo",
    "peerId": "ejemplo",
    "tag": 1,
    "path": [
        "abc",
        "def",
        "cba"
    ]
}
```

---
### Note:
This is done when we are abou to send the message, so the messageModal itself doesn't update the content of the path field

### DIFF
--- a/src/components/Modal/node/SendMessageModal.tsx
+++ b/src/components/Modal/node/SendMessageModal.tsx
@@ -119,7 +119,6 @@ export const SendMessageModal = (props: SendMessageModalProps) => {
     if (sendMode == 'path') {
       const pathElements: string[] = [];
       const lines = path.split('\n')
-      console.log(lines)
       for (const line of lines) {
         const elements = line.split(',').map((element) => element.trim()).filter((element) => element !== '');
         pathElements.push(...elements);
--- a/src/components/Modal/node/SendMessageModal.tsx
+++ b/src/components/Modal/node/SendMessageModal.tsx
@@ -131,21 +131,20 @@ export const SendMessageModal = (props: SendMessageModalProps) => {
       messagePayload.path = validatedPath;
     }
 
-    console.log(messagePayload)
-    // dispatch(actionsAsync.sendMessageThunk(messagePayload))
-    //   .unwrap()
-    //   .then((res) => {
-    //     console.log('@message: ', res?.challenge);
-    //     set_status('Message sent');
-    //     handleCloseModal();
-    //   })
-    //   .catch((e) => {
-    //     console.log('@message err:', e);
-    //     set_status(e.error);
-    //   })
-    //   .finally(() => {
-    //     set_loader(false);
-    //   });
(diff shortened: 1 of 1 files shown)


src/components/Modal/node/SendMessageModal.tsx
```diff
@@ -117,25 +117,35 @@ export const SendMessageModal = (props: SendMessageModalProps) => {
...
-      const pathElements = path.replace(/(\r\n|\n|\r| )/gm, '').split(',');
+      const pathElements = path
+        .split('\n')
+        .map((line) =>
+          line
+            .split(',')
+            .map((element) => element.trim())
+            .filter((element) => element !== '')
+        )
+        .flat();
```
mjadach-iv, line 128: What are the reasons for replacing .replace with map in a map and then flat? Is that more performant?

src/components/Modal/node/SendMessageModal.tsx
```diff
@@ -117,25 +117,35 @@ export const SendMessageModal = (props: SendMessageModalProps) => {
...
+    //     set_status('Message sent');
+    //     handleCloseModal();
+    //   })
+    //   .catch((e) => {
+    //     console.log('@message err:', e);
+    //     set_status(e.error);
+    //   })
+    //   .finally(() => {
+    //     set_loader(false);
+    //   });
```
mjadach-iv: Why is that commented out?

src/components/Modal/node/SendMessageModal.tsx
ItsLuized: The change is not related to performance, more about being able to easlly understand what is happening.
With this change we can follow the though process, (after the map line) about splitting the commas, removing the whitespace, if the element is an empty string, then we don't need it since it is a product of a consecutive comma.

We could do it in regex if it is preffered in this case.
the regex would be something like `.replace(/,+/g, ',').split(',');`

mjadach-iv: Unfetunetely .map inside a map, and split after slit is not easy to read.
Please make it easier to read. If you need help, let me know

mjadach-iv (approved):

PR #451, opened 2023-10-02
Update to a Select for the receiver field

### LINKED ISSUE #444: Message Modal: Combobox with aliases and peer ids instead of texfield
The values on the list should look like:
`alias (PeerId)` if alias is present, and if not, then we should see only PeerId

ItsLuized:
Fixes #444

# Overview

This changes the Textfield that was used to the Receiver, for a Select component, to let the user choose from the peers instead of having to type the peerId. 

It displays the peers as `alias (PeerId)` if an alias exists for that peer, else it shows `peerId`

### DIFF
--- a/src/components/Modal/node/SendMessageModal.tsx
+++ b/src/components/Modal/node/SendMessageModal.tsx
@@ -90,13 +90,7 @@ export const SendMessageModal = (props: SendMessageModalProps) => {
         set_numberOfHops(0);
         set_path('');
     }
-    if (!(loginData.apiEndpoint && loginData.apiToken)) return;
-    dispatch(actionsAsync.getPeersThunk({
-      apiToken: loginData.apiToken,
-      apiEndpoint: loginData.apiEndpoint
-    }))
-    console.log(peers)
-    console.log(aliases)
+
   }, [sendMode, path, numberOfHops]);
 
   const handleSendMessage = () => {
--- a/src/components/Modal/node/SendMessageModal.tsx
+++ b/src/components/Modal/node/SendMessageModal.tsx
@@ -260,11 +260,11 @@ export const SendMessageModal = (props: SendMessageModalProps) => {
         </TopBar>
         <SDialogContent
         >
+          <span style={{ margin: '0px 0px -2px' }}>Receiver:</span>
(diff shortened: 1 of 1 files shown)


mjadach-iv (changes requested):
The task was to use Combobox  and not Select
Example in MUI: https://mui.com/material-ui/react-autocomplete/#combo-box
Example in our code: https://github.com/Rpc-h/degen-rpch-net/blob/94e1bd66a3d940ac0a90b878ef113e47074758bf/client/pages/degen/dashboard.tsx#L276

src/components/Modal/node/SendMessageModal.tsx
```diff
@@ -236,14 +278,28 @@ export const SendMessageModal = (props: SendMessageModalProps) => {
...
         <SDialogContent
         >
-          <TextField
-            label="Receiver (Peer Id)"
-            placeholder="16Uiu2..."
-            value={receiver}
-            onChange={(e) => set_receiver(e.target.value)}
-            required
-            fullWidth
+          <span style={{ margin: '0px 0px -2px' }}>Receiver:</span>
```
mjadach-iv, line 281: that span is not needed as in line 296 you have it already there

src/components/Modal/node/SendMessageModal.tsx
```diff
@@ -68,15 +68,38 @@ export const SendMessageModal = (props: SendMessageModalProps) => {
...
+    multiAddr: string;
+    heartbeats: {
+      sent: number;
+      success: number;
+    };
+    lastSeen: number;
+    backoff: number;
+    isNew: boolean;
+    reportedVersion: string;
+  } | null>(props.peerId ? findReceiver(props.peerId) : null);
```
mjadach-iv, line 99:
```suggestion
  } | null>(props.peerId ?
     peers.find(elem=>elem.peerId === props.peerId )
    : 
    null);
```

mjadach-iv: should be more readable and shorter like that

mjadach-iv (changes requested):
Last changes and we merge

PR #463, opened 2023-10-13
getNodeData

mjadach-iv:
image


src/utils/date.tsx
```diff
@@ -0,0 +1,48 @@
...
+  } else {
+    const d = new Date(epoch);
+    const year = d.getFullYear();
+    const month = d.getMonth()+1 < 10 ? `0${d.getMonth()+1}` : d.getMonth()+1;
+    const day = d.getDate() < 10 ? `0${d.getDate()}` : d.getDate();
+    const hours = d.getUTCHours() < 10 ? `0${d.getUTCHours()}` : d.getUTCHours();
+    const minutes = d.getUTCMinutes() < 10 ? `0${d.getUTCMinutes()}` : d.getUTCMinutes();
+    const formatted = <>{`${year}-${month}-${day}`}{twoRows ? <br/> : ' '}{`${hours}:${minutes}`}</>
+    return <>{formatted}</>;
+  }
```
diegoalzate, line 47: I think we can do this solution with less dependencies with dayjs and probably less lines

relative time: https://day.js.org/docs/en/plugin/relative-time (similar to time ago and we already have the dep)
format time: https://day.js.org/docs/en/display/format

The two rows prop i would move to a special component so the util function returns a string. This being a util func maybe we can even separate it out into a couple functions like:

- formatDateToRelativeTime
- formatDateToUTC
- formatTimeToUTC

we actually already have `formatDateToUserTimezone` and formatTimeToUserTimezone` which can return what you are already returning but in the user timezone maybe you want this in UTC.

After having these seperate util funcs then it would be easy to create a component in the components folder that can even be called TimeAgo doing something like the logic in the if statement that you have there.

```
function TimeAgo(props: { twoRows: bool }) {
// if logic implemented here and chooses which util function to use
}
```

mjadach-iv: Created isssue for that: https://github.com/hoprnet/hopr-admin/issues/464

diegoalzate (approved):

Issue #378, opened 2023-09-08
Formatting v2


diegoalzate:
### Overview
Ok let's try this again. We need formatting in this repo and it shouldn't be a headache.

### What went wrong last time?
- too complicated https://github.com/hoprnet/hopr-admin/pull/56#discussion_r1232357516
- we needed a separate extension for it to actually work
- it seemed to work differently on different machines causing more trouble than it solved.

> IMO this all came down to trying to get more out of prettier than what prettier is normally used for

### What should we do this time?
- simple formatting, get a prettier config from another project and a eslint config from another project and make sure they are used separately
- possible inspirations: `rpch` and `hopr-sdk` 
- just use prettier, if prettier can't do it then remove it
- update scripts for formatting and fixing formatting

Simple examples:
- https://github.com/vercel/next.js/blob/canary/.prettierrc.json
- https://github.com/trpc/trpc/blob/main/.prettierrc.js
- https://github.com/vuejs/vue/blob/main/.prettierrc

we already have a simple prettier https://github.com/hoprnet/hopr-admin/blob/main/.prettierrc.json we just need to make eslint do less

PR #466, opened 2023-10-16
add lint staged

diegoalzate:
closes https://github.com/hoprnet/hopr-admin/issues/378

### overview
Formats on commit but only staged files, no unexpected formatting.

### possible improvements
lint on commit, right now linting can be seen by running yarn lint:fix, adding linting to this pr can increase the complexity quite a bit and also may slow down dev speed by enforcing rules that in some cases has allowed to deploy in time. I would recommend adding it but after this pr.

inspired by https://prettier.io/docs/en/precommit


ItsLuized (approved):
I really like this approach to formatting!

PR #479, opened 2023-10-29
Update import of aliases to use async-await

### LINKED ISSUE #471: Import CSV issues
Expectations: When trying to export aliases as a csv file and whenever aliases page is empty (either aliases are deleted manually or node has been restarted), and trying to import from previously generated CSV file it should import all the aliases.

Current issue: When trying to import csv file, it just grabs first like and it doesn't go over the whole file.

ItsLuized:
Fixes #471 

# Overview

This function didn't have async await, to result of this was all the request being executed at the same time, since we have a condition where if a function is called while it's already executing, we were cancelling before it's execution (refer to: https://redux-toolkit.js.org/api/createAsyncThunk\#canceling-before-execution).

When using await to lock the execution to not be parallel it works as expected

## Demo Result
Screen Recording 2023-10-30 at 06 27 06

### DIFF
--- a/src/pages/node/aliases.tsx
+++ b/src/pages/node/aliases.tsx
@@ -85,10 +85,10 @@ function AliasesPage() {
   };
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const handleCSVUpload = (parsedData: any[]) => {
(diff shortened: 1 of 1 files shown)


diegoalzate (approved):
good find!

mjadach-iv (approved):

PR #482, opened 2023-11-07
Multisig option for transactions

### LINKED ISSUE #461: Multi signature on Staking Hub UI
- add option to add new signers on the dev page
- add option to set the threshold on the dev page
- add UI for multisig (similar to safe ui) in all the places where we need to sign sth
- if treshold is higher than 1, give options like there are on the safe ui
- only let other users to sign the actions on the transations page (not be switching the wallet address and staying on the same view to sign)

diegoalzate:
closes #461 

### overview
- create SafeTransactionButton that handles signing or executing tx
- implement SafeTransactionButton with side effect changes in relevant places
- some renaming of thunk actions so they are more consistent

### disclaimer
- feedback for signing is lacking, since there is no transaction hash the feedback component does not know if the signing was a success, a possible improvement on this is adding safeHash as a prop to the feedback component and querying the indexer to mimick useWaitForTransaction but with safe.
- ux on onboarding for signing needs to be revisited, since execution is done on the transactions page this means that right now you can sign your multisig safe but there is no way to know when it was actually executed from that same page so a refresh might be required and might not be the best solution.


mjadach-iv:
Merged from a different branch

PR #484, opened 2023-11-15
Node/update 2.0.1

mjadach-iv:
Changelog:
- update HOPRd SDK to 2.0.1
- add node actions such as ping, add alias, send message to channels page
- disable ping, add alias, send message modals on channels page if we do not know the peerId of the node
- update ping, add alias, send message modals to update the peerid in the textfield if it uodates (long load times can happen) 
- add copy and 'see on gnosisscan.io' buttons on important eth addressis on the info page
- add wxHOPR allowance to InfoBar
- remove 'redeem tickets button' from `/tickets` page
- add wxHOPR parsing to `/tickets` page
- remove pending from  `/tickets` page

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -1,4 +1,4 @@
-export const environment: 'dev' | 'node' | 'web3' = 'dev';
+export const environment: 'dev' | 'node' | 'web3' = 'node';
 
 // Smart Contracts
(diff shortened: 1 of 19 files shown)


esterlus (approved):
Please evaluate if it might be better to remove the commented code instead

Issue #478, opened 2023-10-27
Node Admin improvements


mjadach-iv:
- [x] Hide strategies
- [x] If possible to access address in gnosisscan, put a url link there
- [x] hoprToken, hoprChannels copy imrpove
- [x] channels node address to gnosisscan
- [x] send message from chanels page
- [x] color xDai if below treahhold
- [x] failed to add alias - error toast
- [x] message timestamp column, wrap messages, 
- [x] Color the status in right bar
- [x] Redownload balance on Node Admin refresh, as it is not downloaded if you are not in the Info tab

mjadach-iv:
some parts of the issue moved here as they require store redesign: https://github.com/hoprnet/hopr-admin/issues/485

PR #486, opened 2023-11-21
Ux/improvements

mjadach-iv:
Changelog:
- load all info to infobar on page refresh
- color status and if not enough balance - also color the native coin
- peer quality visual representation in table peers
- count total wxHOPR staked and display that sum in Info subpage and InfoBar component
- show error on adding aliases
- add timestamp to received messages and wrap long messages in the table

Closes https://github.com/hoprnet/hopr-admin/issues/478

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hopr-admin",
-  "version": "0.6.7",
+  "version": "0.6.8",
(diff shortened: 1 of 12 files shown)


Jaguaras (approved):
lgtm!

PR #496, opened 2023-11-28
Node PeerID copy button and remove web3 login froe Node Admin mobile

mjadach-iv:
Changelog:
- Node PeerID copy button on the info page
- remove web3 (Staking Hub) login from Drawer Node Admin mobile


Jaguaras (approved):
lgtm

PR #505, opened 2024-01-11
Docker Command in Nodes

mjadach-iv:
Docker Command for running Docker Node from the Dashboard
The cmd is imported from the onbloarding cmd

image


Jaguaras (approved):
amazing!

esterlus (approved):
formatting not consistent though, forgot to run prettier/eslint?

e.g.: `(moduleAddress?: string | null)` vs `(moduleAddress?: string| null)`

PR #511, opened 2024-01-22
bugfix/rounder

mjadach-iv:


esterlus (approved):
approved because of related issue to improve this

PR #537, opened 2024-02-13
Staking Hub for 2.1

mjadach-iv:
- change onboarding to add annoucement contract
 - update old safe button to add annoucement contract
 - add over-the-page loaders to critical pages
 - make onboarding loader on on load of the page and turn it off when we have all onboarding data
 - make more use of subgraph instead of safe infra
 - mobile view for dashboards - making elements responsive
 - add option to load safe and module from url without owner in the wallet
 - do not load last used wallet address on page load if wallet is offline
 - fix 'on/off' bahaviour of tran icons for main and next onboardigns on the dashboard/node page
 - hide redemeed tickets from dashboard


Jaguaras (approved):
Looking great!

PR #550, opened 2024-02-29
Fix node onboarding train icon and other minor stuff

mjadach-iv:
- Fix node onboarding train icon
 - remove unused graybox for Earned rewards
 - fix .includes() error on nextNode onboarding


esterlus (approved):
would be even better with consistent formatting

PR #563, opened 2024-03-21
Fix checksums on ETH addreses and better error showing

mjadach-iv:
Needes full testing on:
- withdraw subpage
- fund node subpage
- edit owners subpage


Jaguaras (approved):
Looking great!

